PR #22, opened 2015-02-16
Merge Android port branch into master

davtbaum:
PR for merging android port into master. It's pretty clean as most of the meat is in the toolchain file and few bugs I found here and there. The README in ./android covers the building process.

@rpavlik, one thing that needs to be addressed is reverting back from cross compilation when building the devtools osvr_json_to_c binary. My hacky solution was to create the devtool in a separate native build, then  pull it in during cross compilation. I wanted to consult with you first on the best way to do this. Some research so far that may be helpful:
http://www.vtk.org/Wiki/CMake_Cross_Compiling#Using_executables_in_the_build_created_during_the_build


rpavlik:
OK, so at this point I think we'll probably just piecemeal cherry-pick/rewrite bits of this as it goes over, since that will be both easier and cleaner than a direct merge. Will try to split up things you found/learned into discrete issues.

rpavlik:
OK, the only thing we haven't merged yet (or subsumed into https://github.com/OSVR/OSVR-Android-Build ) is the readme file you added, so I've filed an issue about it there https://github.com/OSVR/OSVR-Android-Build/issues/4 with a copy of the file in a gist, so I can close out this old PR. Thanks!

PR #39, opened 2015-03-08
Use an HTTPS URL for the JSON Schemas submodule.

Armada651:
It is more consistent and easier to use.


rpavlik:
I definitely agree - happy now that this is public, switching to the HTTPS URL won't confuse the CI.

Thanks for your contribution! Happy to have you onboard!

PR #86, opened 2015-04-14
Merge USB Serial Autodetect/Enumeration lib

mars979:


rpavlik:
So this needs a Linux port in order to pass CI. It should be pretty easy, since I think you can just use sysfs to do it (so it turns into a filesystem traversal), but that will block the merge. Would like to get this merged ideally shortly before 0.2, since changes to the manual configuration API will be coming in 0.2 which directly affects this code's usage.

@yboger Is this a reasonable item to have @godbyk do?

PR #93, opened 2015-04-24
Updated JSON display configs.

DuFF14:
Moved updated display configs from Unity repo.

### DIFF
--- a/apps/displays/HMD.json
+++ b/apps/displays/HMD.json
@@ -1,34 +1,41 @@
 {
     "hmd": {
+        "device": {
+          "vendor": "OSVR",
+          "model": "HDK",
+          "Version": "1.1",
+          "num_displays": 1,
+          "Note": "Settings are also good for HDK 1.0"
+        },
         "field_of_view": {
             "monocular_horizontal": 90,
             "monocular_vertical": 101.25,
             "overlap_percent": 100,
             "pitch_tilt": 0
         },
-		"resolutions": [
-			{
-				"width": 1920,
-				"height": 1080,
-				"video_inputs": 1,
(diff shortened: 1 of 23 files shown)


rpavlik:
So a few notes:

- Key names are inconsistent in style: typically json uses camelCase because that's the Javascript convention. You've got a mix of initial caps and snake_case.
- Unclear whether the version field refers to version of the config file or version of the hardware. In the device descriptor it's version of the config file.

I'll make a file with general updates elsewhere - those are just the changes I noticed in this pull request.

rpavlik:
Oh, and you may want to use Atom or a similar web-savvy editor for editing JSON

DuFF14:
I wasn't sure about Json version number because the existing configs left that field empty or have something like "landscape". I am installing Atom so I will go through and make things consistent. I figured since version is in the "device" field that it referred to the hardware version.

Should there be a config(s) for Oculus in this directory, or does that belong in the Oculus repo?

yboger:
Yes, there should be a config for Oculus.

Yuval

rpavlik:
but the display descriptor, just like the device descriptor, belongs in the Oculus repo - since the config file that would refer to it is there, and eventually that plugin will have it embedded within it.

rpavlik:
Since this format is in flux, can we get a schema version number in here?

DuFF14:
Added schema version number.
These files would be parsed correctly in the DisplayConfig Unity branch: https://github.com/OSVR/OSVR-Unity/pull/41

rpavlik:
I'm going to close this request, since the non-cosmetic changes were just merged in a different pull request.

PR #96, opened 2015-04-28
An external VRPN example using a Polhemus G4 system with one sensor and ...

JermAker:
An external VRPN example using a Polhemus G4 system with one sensor and one PowerTrak 360 sensor with buttons.


rpavlik:
Looks great, thank you for your contribution!

PR #97, opened 2015-04-30
Modified the reset-yaw application so that it applies the transformation...

russell-taylor:
... as a post-rotation.  This causes the HMD motion to remain correct for local rotations, so that we don't scramble the axes.


russell-taylor:
The fix to avoid axis scrambling after rotation applied here has been tested on a pre-0.2 release and worked.  The app currently does not compile (but it also does not compile on master).

rpavlik:
Will forward-port this to the post-0.2 master.

rpavlik:
Almost done with the port. Cleaned up and deduplicated things as I went.

rpavlik:
OK, this is just about ready. Think I need to quick fix https://github.com/OSVR/OSVR-Core/issues/134 to make it robust, since it may add a transform at a different alias level than an existing transform which would result in unexpected behavior, but it does work tested here with a YEI tracker using the HDK YEI config file.

rpavlik:
The updated branch includes a fix for #134.

yboger:
Need to publish API so that game developers can execute this function from the game

rpavlik:
As I mentioned to Greg the other day:

> My concern is that doing this is actually doing something more fundamental inside the logical structure: it's changing the calibration of coordinate systems, so if you have position tracking, other trackers, etc. you might not want to accidentally screw up that calibration because a game decides it always starts facing forward.
>
> Now, if a game wants the user's facing direction at startup to be forward, they can do that themselves - by applying a transform to their game world - without breaking calibration

The reset-yaw persists as long as the server is running - if we want to make the transform persist between restarts we can add that. (Apparently every time you start a game with Oculus it realigns on every game launch, so the "recenter" is needed more there since they aren't keeping consistent coordinates. Will be interested to see how they deal with that with CV1 now that they have other tracked devices)

We could potentially provide helper APIs for the latter suggestion in the quotation (adjusting the world to room transform), but doing that in the transform config of the server is really an unsutainable hack in most cases.

Particularly with positional tracking, where we'll now have another coordinate system coming in, me being somewhat of a stickler on this point will become more important. If we determine that the camera defines a fixed room coordinate system, we can certainly make it so that the transform from optical tracker space to IMU tracker space is reset that way so the "reset yaw" becomes permanent and linked to the camera, which probably makes good sense. We just really want to make sure we achieve the results we want by adjusting transforms at the right level or we'll be in a world of hurt.

yboger:
Right, but many game engine people don't know (or don't have the time) to do this. So, I would still like to see API to reset yaw from a game. 

Good or bad, game developers are used to having this call for the Oculus API so they are expecting it here as well.

We could provide documentation that explains the concerns that this might have a bigger impact,

rpavlik:
OK, will open a separate issue for this. Maybe the API we provide would do a global transform - transform all trackers - instead of the single sensor transform this code does.

PR #103, opened 2015-05-07
Added tests for header dependencies. Patched broken headers.

godbyk:


inc/osvr/Util/ResetPointerList.h
```diff
@@ -44,7 +44,7 @@ namespace util {
         typedef typename RangeType::type IteratorType;
         typedef typename IteratorType::value_type PointerType;
         for (auto &ptr : range) {
-            ptr.reset()
```
rpavlik: pretty sure this header isn't used anywhere, but that's a separate issue.

godbyk: It isn't; I checked. I figured I'd fix it anyway. I wasn't sure if you have future plans for it or not.

tests/header_dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+        $<BUILD_INTERFACE:${BUILDTREE_HEADER_BASE}>
+        $<BUILD_INTERFACE:${HEADER_BASE}>
+        $<INSTALL_INTERFACE:include>)
+
+    if (${extension} STREQUAL cpp)
+        target_compile_options(${test_name} PUBLIC ${OSVR_CXX11_FLAGS})
+    endif()
+
+    target_link_libraries(${test_name} ${COMMON_TEST_LIBRARIES})
+    if (LIBRARIES_${basename})
```
rpavlik: is `basename` ever defined anywhere? Looks like a copy-pasteo.

godbyk: @rpavlik See lines 18–19.

rpavlik:
Would be good to put all these millions of targets into a project folder: see for example https://github.com/OSVR/OSVR-Core/blob/master/apps/CMakeLists.txt#L17

tests/header_dependencies/CMakeLists.txt
rpavlik: ah, right on.

tests/header_dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+
+file(GLOB_RECURSE headers RELATIVE "${HEADER_BASE}" ${HEADER_BASE}/*.h)
+foreach(HEADER ${headers})
+    list (FIND IGNORED_HEADERS "${HEADER}" _index)
+    if (${_index} GREATER -1)
+        continue()
+    endif()
+
+    string(REPLACE ".h" "" basename "${HEADER}")
+    string(REGEX REPLACE "[\\./]" "_" basename "${basename}")
```
rpavlik: This step might be made easier/clearer/better by `string(MAKE_C_IDENTIFIER`, maybe? http://www.cmake.org/cmake/help/v3.0/command/string.html

godbyk:
@rpavlik I added a commit to put the targets in a folder.

tests/header_dependencies/CMakeLists.txt
godbyk: @rpavlik As long as C identifiers are valid CMake variable/target names, I'm cool with that. Saves me having to fiddle with that regex if we create an oddly-named header file.

rpavlik:
Looks nice, thanks! Might find one or two more issues in CI, we'll see.

PR #112, opened 2015-05-08
Ensure that forward-declaration headers are included by full-declaration headers

godbyk:
Added tests to sure that `_fwd.h` and `Ptr.h` headers are included in their primary header files.

Running `ctest -R test_fwd_headers` will execute this test.

Fixed headers that failed this test.


godbyk:
I hope Windows CMake supports `break()`.

rpavlik:
Should be good to merge, esp. since it doesn't change the build, just the test runs.

PR #115, opened 2015-05-11
Add example configs for external VRPN devices

mars979:


apps/external-devices/osvr_server_config.externalvrpn.ART_DTrack_Flystick.json
```diff
@@ -0,0 +1,43 @@
...
+          "tracker": {},
+          "button": {},
+          "analog": {}
+        }
+      }
+    }
+  },
+  "aliases": {
+    "/me/head": {
+      "rotate": {
```
rpavlik: if this is just a null/identity transform, it can just be `"/me/head": "/DTrackDevice/tracker/0"`

rpavlik:
OK, there are a few changes for the future, but this doesn't break anything, so let's get it in.

apps/external-devices/osvr_server_config.externalvrpn.ART_DTrack_Flystick.json
mars979: There's no need to transform since it's using the powerwall coordinate system. But I left it as an example, in case they use a different one and will need an example how to perform transforms

rpavlik: OK, that seems reasonable, esp. since I didn't really write any docs on how to write transforms.

PR #122, opened 2015-05-13
Android 3

DuFF14:
Added OSVR_ANDROID platform definition and handle OSVR_ANDROID case in USBSerialDevInfo.cpp identically to OSVR_LINUX. USBSerialDevInfo_Android.h is a duplicate of USBSerialDevInfo_Linux.h, except instead of including "linux/serial.h" (because it could not be found in Crystax NDK), it includes a new file (USBSerial_Android.h) which contains the necessary definitions from serial.h.


rpavlik:
I don't think we can do this license-wise, copying in that code.

DuFF14:
The comments say this file has no copyrightable information: https://android.googlesource.com/platform/development/+/73a5a3baaa5089f1ab2049e5934fa5d8a3f3e76a/ndk/platforms/android-20/include/linux/serial.h

serial.h exists in the NDK but not until android-20, so this relates to the discussion here: https://github.com/OSVR/OSVR-Android-Build/issues/6

An alternative could be building for android-21 until we have a solution for earlier platforms, but when I tried that initially it caused the build to break while building vrpn server.

rpavlik:
Well, that's google's assertion, I don't recall if they won that court case or not. In any case, it shouldn't be imported directly into a file with Sensics Apache 2.0 boilerplate, and it's probably not a good idea to rely on it if the underlying functionality is not stable/guaranteed to be there (which is effectively what they're saying by not having the header).

The alternative not mentioned is to disable the usbserial on android and conditionally-compile the YEI driver's usage of it.

DuFF14:
Build successful after disabling USBSerial and OpenCV Capture plugin, so these are two remaining issues.
This is the OpenCV error I'm stuck on: https://gist.githubusercontent.com/DuFF14/fdfe4ecf833111ffb100/raw/04b60cfef071f4b8c5f3e2be9fa1185f172ad2ea/gistfile1.txt

rpavlik:
OK, I've rebased and enhanced this code, got a build that at least appears to run.

PR #131, opened 2015-05-26
Shared memory

rpavlik:
This is an initial iteration with shared memory support for imaging. Enables larger images and multi-byte depth per channel, faster (because of using shared memory vs serializing into/out of network buffers).


rpavlik:
@godbyk thanks for the code review. I had forgotten to try building here with gcc/mingw, and further with Linux - we really need to get a pull request builder set up properly.

Remaining issues:
- build error in Imaging.cpp example under mingw64 gcc
- what to do with `xsi_shared_memory`
- rename class and probably split up files - that `.cpp` file is much longer than I'd like.

rpavlik:
OK, have addressed the issues that came up, and now have it building in my own tests on MSVC 2013, gcc mingw64, and debian jessie gcc.

Issue #132, opened 2015-05-27
Clean up, import updated display descriptors


rpavlik:
This is without changing the schema - should import the VRVANA descriptor, and back-port changes from others (found in the Unity plugin?)

@DuFF14 @yboger Here is a json schema for the descriptors currently in OSVR-Core master - is this the format you want me to be keeping here? https://gist.github.com/rpavlik/ae6b2d033d804be81156 (It's human-readable)

DuFF14:
If the goal is to update the display descriptors without breaking parsing as it
is now, looks like this will work.

rpavlik:
OK. Yes, that is the current goal. Then I will make the requested changes keeping that schema for now.

rpavlik:
Corresponding pull request merged.

PR #133, opened 2015-05-28
Added updated display descriptors

rpavlik:
This solves #132 - I grabbed all the ones I could find and converted them (scripted, combo of `jq` and `sed`).  Note that we're right now splitting up the multi-resolution descriptors into single-resolution ones.

@yboger @DuFF14 If I could get a review of these and a clear to merge, that would be great.


DuFF14:
Unity parser expects an integer for the rotate_180 field right now. Mentioned here as well: https://github.com/OSVR/OSVR-Unity/pull/46 
Some of these have an integer for this field (Vrvana, HMD.json), others have "true" or "false" (HDK, DK2). That is the only issue I noticed.

rpavlik:
Hmm, so I assumed that it would do the conversions and that you were reading them as bools. I think it was my script that added them as "false" if they didn't already exists. Quite awkward as an integer, but since we're keeping the schema for now I can update that.

rpavlik:
OK, re-ran the conversion script after changing the default to 0, and I've updated the schema, will be putting it in the repos. Can you test?

DuFF14:
Vrvana is missing swap_eyes, num_displays, and note. Are these optional? Would default to swap_eyes=0 and num_displays = 1.
Tested with HDK and DK2. Looks good to merge.

rpavlik:
If you want me to add those to all, I can, but I'd prefer to just let those default, particularly since num_displays is not long for this world...

rpavlik:
And what is the `HMD.json` file? It looks like the HDK with less detail.

DuFF14:
Yea I don't think it's necessary, just a template or default.

DuFF14:
I don't have an issue with letting them default.

rpavlik:
OK, will merge. I've pushed the associated schema to the schema repo and updated the submodule here too.

Issue #127, opened 2015-05-22
Support for remote OSVR servers 


janoc:
I am attempting to use OSVR with a Samsung Gear mobile HMD. Our use case is a client running on the HMD, connecting to another machine that hosts all the devices the application may need - controllers, trackers, navigation devices, etc. There will be nothing connected directly to the HMD because it is impractical, thus there is no need for an OSVR server to run on the headset - reducing complexity and saving battery power.

Right now the high level API doesn't support this use case. The class `osvr::clientkit::ClientContext` doesn't allow to specify the host where the OSVR server is located. This can be worked around either by using the low level API (which does take a `host` argument) or by using an environment variable. Neither workaround is ideal - the low level API exposes Boost dependencies to the client application, making for a complicated compilation, especially on mobile platforms. The environment variable approach is not usable on Android.

##### Proposed enhancement
Add an additional constructor to the `osvr::clientkit::ClientContext` class,  the `host` argument accepting also the `host:port` format (important for servers running on non-standard ports):

```c++
inline ClientContext::ClientContext(const char applicationIdentifier[],
                                                      const char host[],
                                                      uint32_t flags);
```

rpavlik:
Or, alternately, come up with a non-hacky alternative to the environment variable.

(I think `host:port` is already supported transparently with the APIs that accept a host)

janoc:
Yes, it is - the low level API accepts that and it should even work as vrpnConnection is used underneath.

janoc:
Updated the pull req for the latest master: https://github.com/OSVR/OSVR-Core/pull/331

janoc:
Hello, 

Is there a chance to get this trivial enhancement merged? I have encountered this again when updating our in-house OSVR copy today. The pull request is over 2 years old and doesn't conflict with anything. Right now we have to either use our in-house built copy or use a kludge with setting the environment variable from inside of the application, which is ridiculous. 

If it isn't going to be merged, then I would at least like to know why so that I could accommodate that locally.  

Thank you.

Issue #139, opened 2015-06-04
OSVR Remote clients fail with internal devices


janoc:
OSVR server advertises its local devices (e.g. a Razer Hydra) in the tree as `"com_osvr_Multiserver/RazerHydra0@localhost"`. 

This causes remote clients to attempt to connect to "localhost" - which is obviously not the place where the device is being hosted, resulting in the UDP connection from the client failing. 

@rpavlik suggested a fix by doing a search & replace of `localhost` in the received tree by the actual host string used for the connection. It cannot be done reliably on the server ahead of time because on a multihomed (e.g. wifi + ethernet) machine the client connection could arrive from any of the available interfaces. 

This is related to #127

russell-taylor:
VRPN can be configured to listen on all interfaces.  In fact, it does this by default.  This was modified so that Windows security wouldn't complain about trying to provide network access when only exposing to localhost.

janoc:
@russell-taylor That is actually not the issue (and is good that VRPN does that). However, you could have a client connecting over e.g. ethernet and attempting to use a device that is advertised in the tree as on the wifi interface - which may not be accessible to that client. That's why the `localhost` rewriting should be done on the client - the client knows which hostname/address it used to connect to the server.

rpavlik:
@russell-taylor This is strictly an OSVR-specific bug, not related to VRPN.

PR #142, opened 2015-06-05
Fixes for issues #127 & #139

janoc:
These two commits fix issues #127 and #139

### DIFF
--- a/inc/osvr/ClientKit/Context.h
+++ b/inc/osvr/ClientKit/Context.h
@@ -46,6 +46,11 @@ namespace clientkit {
                                         uint32_t flags)
         : m_context(osvrClientInit(applicationIdentifier, flags)) {}
 
+    inline ClientContext::ClientContext(const char applicationIdentifier[],
+                                        const char host[],
+                                        uint32_t flags)
+        : m_context(osvrClientInitHost(applicationIdentifier, host, flags)) {}       
+        
     inline ClientContext::ClientContext(OSVR_ClientContext context)
         : m_context(context) {}
 
--- a/inc/osvr/ClientKit/ContextC.h
+++ b/inc/osvr/ClientKit/ContextC.h
@@ -63,6 +63,21 @@ OSVR_CLIENTKIT_EXPORT OSVR_ClientContext
 osvrClientInit(const char applicationIdentifier[],
                uint32_t flags OSVR_CPP_ONLY(= 0));
 
+               
+/** @brief Initialize the library.
+
+    @param applicationIdentifier A null terminated string identifying your
+   application. Reverse DNS format strongly suggested.
+    @param host A null terminated string identifying host with the server to connect to.
+    @param flags initialization options (reserved) - pass 0 for now.
+
+    @returns Client context - will be needed for subsequent calls
+*/
+OSVR_CLIENTKIT_EXPORT OSVR_ClientContext
+osvrClientInitHost(const char applicationIdentifier[],
+                   const char host[],
+                   uint32_t flags OSVR_CPP_ONLY(= 0));
+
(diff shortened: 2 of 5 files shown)


rpavlik:
Note: should check to ensure that `m_host` is not empty before doing this, though I think m_host being non-empty is a class invariant (don't know if it's documented or tested as one though)

janoc:
I have tested the case when no host is provided (the default use case with collocated server & client) and it seems to work fine.

rpavlik:
Thanks. Trying to figure out if we put the extra constructor method in a separate library or just say not to use it in the docs, since I don't really think we want game engines using that constructor, but that's all that is blocking the merge

janoc:
Um, why you don't want the game engines use that constructor (the one supplying the host name)? That is a very common use case. If OSVR is meant to be only for localhost and any other use cases unsupported, then *that* is actually going to be a problem.

rpavlik:
Well, it's not meant only for localhost. One of the initial design goals was to keep system configuration/implementation out of the realm of concern of the client application. Having a constructor with a host in it exposes underlying implementation details, which I understand is useful for some use cases (particularly yours :D ) but I'm not sure we should commit to having that in the stable API yet. Not having it means we can do things like a server autodetection (zeroconf/bonjour, etc) and so on behind the scenes and have all client apps pick up on it.

I was thinking that we might have a companion library to ClientKit - maybe an ClientKitX - containing experimental APIs that aren't expected to be as commonly used and not the same api/abi stability efforts we'll have elsewhere, and that would be a decent place to try it out.

@yboger - do we have any game studio feedback on this point? (regarding configuration)

janoc:
I quite fail to see how is a hostname/port a configuration - when I want to visit this website, I have also type github.com at some point (or use a bookmark, but that's a detail). Autodiscovery using Zeroconf or whatever cannot replace that. Honestly, I have yet to find a network-based application that does not allow specifying this somewhere.

IMO, a library should never impose design decisions like this on applications. It is up to the application whether or not they will always use the constructor without the host, constructor with the host and feed in "localhost" every time, load the host name from a configuration file/resource or use some sort of auto discovery mechanism or whatever. It is the application developer who knows best what their application needs, not the library/middleware vendor. The library is here to provide the tools I need to do my job, not to tell me how to manage my configuration system (which is a completely unrelated issue to managing input in an application). 

You can provide/support some recommended strategies and best practices, but that's all that the middleware should do. Libraries that try to impose policies like this are an enormous pain to use in the end, requiring hacks and workarounds to get the work done. 

You cannot anticipate every situation and need and are just constraining the usability of the library to a specific usage pattern. The mere existence of this issue here is actually the best evidence of that.

janoc:
And re having a separate library - I could have actually implemented the change exactly like that. It was my first idea how to do it, so that I don't have to modify OSVR. It would have been also faster, because I wouldn't have needed to compile the Core. 

The reasons why I didn't do it like that in the end:
* It is an extra library that has to be loaded, deployed and consistency managed, especially as it needs low level/non-public APIs. That's a pain when dealing e.g. with Android, .NET wrappers, Unity, etc.
* Consistency - it would need a factory function creating the ClientContext when the hostname is needed. When it isn't needed, constructor can be used. Really ugly.
* It is a logical and straightforward extension of the existing ClientContext, especially when the low level code provides the functionality already.

rpavlik:
Can you split this into two pull requests? (I could cherry-pick, but then we don't get a clean merge commit, etc) The localhost fix is a definite keeper, but I'd like to think/discuss some more on the other patch.

janoc:
OK, here you go: 

Hostname enhancement #127 
https://github.com/OSVR/OSVR-Core/pull/154

Bugfix for the @localhost problem #139 
https://github.com/OSVR/OSVR-Core/pull/152

PR #152, opened 2015-06-18
Fix for issue #139

### LINKED ISSUE #139: OSVR Remote clients fail with internal devices

janoc:
Implemented localhost rewriting (fixes #139)

### DIFF
--- a/src/osvr/Client/PureClientContext.cpp
+++ b/src/osvr/Client/PureClientContext.cpp
@@ -34,9 +34,11 @@
 #include "ButtonRemoteFactory.h"
 #include "ImagingRemoteFactory.h"
 #include "TrackerRemoteFactory.h"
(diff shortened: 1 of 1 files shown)


rpavlik:
Thank you!

PR #154, opened 2015-06-18
Added API for remote clients

janoc:
Fixes issue #127, discussion here: https://github.com/OSVR/OSVR-Core/pull/142


rpavlik:
(I do actually plan on including this at some point, btw...)

janoc:
Thanks. I wouldn't have mentioned that but we had to update OSVR here because of Unity so now I am busy updating and recompiling ...

PR #156, opened 2015-06-18
Video based tracker

rpavlik:
Needs some porting to 0.2, and some cleanup (particularly some Eigen usage).


rpavlik:
OK, I rebased this branch on the latest master, fixing the (primarily build system) conflicts. I also fixed up the files a smidge (re-wrapping comments, pasting the headers into the templates from the boilerplate generator which I really need to port to Ruby and stick on a better URL). It builds, haven't tested it.

rpavlik:
OK, now have a standalone app that can do camera tracking, next will abstract out the source of the images so we can have an automated test run on the saved images.

rpavlik:
OK, so I did a fair bit of cleanup, though didn't get it fully testable yet. Thinking we can probably merge this into master (it currently will build as part of CI builds, but the plugin doesn't get installed) - I tested the build on Windows MSVC and MinGW/GCC 4.9.2. Have a few questions that I'll stick in as comments on the pull request.

plugins/videobasedtracker/HDKLedIdentifier.cpp
```diff
@@ -0,0 +1,135 @@
...
+        // the threshold computed above.
+        auto bits = getBitsUsingThreshold(brightnesses, threshold);
+
+        // Search through the available patterns to see if the passed-in
+        // pattern matches any of them.  If so, return that pattern.  We
+        // need to check all potential rotations of the pattern, since we
+        // don't know when the code started.  For the HDK, the codes are
+        // rotationally invariant.
+
+        /// @todo feels like there should be a good algorithm for
```
rpavlik: I feel like this is something we covered in my undergrad CS curriculum - string matching/alignment, might have been related to dynamic programming - anyway, a std::list<bool> is terribly inefficient esp. when the processor should be able to do this at most with some rotates and compares.  Any thoughts on the algo I can't think of offhand here, @russell-taylor ?

plugins/videobasedtracker/Types.h
```diff
@@ -0,0 +1,81 @@
...
+    typedef std::list<Brightness> BrightnessList;
+    typedef std::pair<Brightness, Brightness> BrightnessMinMax;
+
+    typedef std::unique_ptr<BeaconBasedPoseEstimator> EstimatorPtr;
+    typedef std::unique_ptr<LedIdentifier> LedIdentifierPtr;
+
+    typedef std::list<Led> LedGroup;
+
+    /// @name Containers of "per-sensor" objects
+    /// @brief It seems like in a "well-formed" video-based tracker, there is
```
rpavlik: Did I correctly deduce this invariant? @russell-taylor

plugins/videobasedtracker/HDKLedIdentifier.cpp
russell-taylor: I could certainly believe that there is a more efficient way to do this.  The system was operating at full video rate on my laptop and there are features to add, so I wasn't trying to optimize every step.  Whether the more-efficient way is easy to understand as working properly I don't know, so I recommend holding off until we have regression tests on the subparts.  Does profiling show this to be a large fraction of the overall time?  I would expect the video processing to be taking the lion's share of the time.

plugins/videobasedtracker/Types.h
russell-taylor: I think so.  The obvious violation in Oculus code happens because the commented-out parts of the code were lost along the way when refactoring.  The original code had both an empty group for the m_led_groups and notes for adding the others:
            // TODO: Fill these in when they are known
            // m_identifiers.push_back(XXX);
            // m_estimator.push_back(XXX);
            std::list<osvr::vbtracker::Led> empty_led_group;
            m_led_groups.push_back(empty_led_group);

rpavlik: OK. I did use those lines to guide my thinking, I just wasn't sure why only two of the three were commented out. I'd think then that the "struct of vectors" should be a "vector of structs" - or rather, that sensor-specific code and data be factored out and combined.

russell-taylor: Its current state is a combination of what was needed to be able to hand to OpenCV, pushing to get something working, and having abandoned the Oculus code in the middle of the project when I realized that we weren't going to be able to configure its camera for synchronous operation.  The Oculus code works enough to get the camera data and keep the DK2 alive (which turns out to be required to do my DirectMode development, so we may want to pull that code out separately).

My guess about being commented out is that only one of the three was needed to make the code run.

russell-taylor:
Fixed the constructors when using the fake images to use the factory instead.  Then it runs with the fake images from the OSVR HDK and displays images correctly.  I think this is ready to merge.

russell-taylor:
Note that this branch needs either boost 1.56 or the patch described for 1.55 in this post: https://svn.boost.org/trac/boost/attachment/ticket/9332/vc12_fix_has_member_function_callable_with.patch

PR #157, opened 2015-06-18
External devices

rpavlik:
@mars979 I see you have new commits here? Is this ready to merge?


mars979:
Sorry, missed your comment. I have a few more devices that need to be added, but we can merge these right now.

PR #158, opened 2015-06-18
Factoriadic number system for universal endianness identification

rpavlik:
Creating a pull request for this, though it'll need some revision before/if it gets merged. The idea was to use factoriadic numbers to uniquely identify arbitrary byte orders, to be able to package a byte order identifier with imager data on the network for remote byte order change if required.


rpavlik:
closing this for now and taking it back to my fork.

PR #159, opened 2015-06-19
Test multiple contexts

rpavlik:
Sequential and simultaneous are in separate executables, to ensure
no (presumably external, VRPN) global state pollutes the test.

All tests pass.


godbyk:
+1 The tests look good to me.

PR #160, opened 2015-06-22
Corrected/safer localhost rewriting, avoiding an undefined behaviour …

janoc:
Corrected/safer localhost rewriting, avoiding an undefined behaviour corner case with std::string:replace()


rpavlik:
Thanks!

Issue #116, opened 2015-05-11
DLL files should have version resources


rpavlik:
We're computing the version from Git - we should be generating a resource file (for Windows) that actually uses that version to put Windows-suitable version info in the dll files, if for no other reason than disambiguation.

I consider this a bug because they're not really "good" dlls without this metadata.

rpavlik:
The resources files should be generated by CMake from the known version information acquired from Git - the version string should contain the "git describe" version including the commit hash.

References:
- http://stackoverflow.com/questions/6693100/how-to-generate-windows-dll-versioning-information-with-cmake
- https://msdn.microsoft.com/en-us/library/windows/desktop/aa381058%28v=vs.85%29.aspx
- https://github.com/OSVR/OSVR-Core/blob/master/cmake-local/Version.cmake

yboger:
This also has benefit for the users that can determine the version of the plugins either directly or through some configuration utility

rpavlik:
Yes - was working on this Monday, trying to get it done for the updated Unity plugin release. I have it mostly done, just troubleshooting why `rc.exe` (in visual studio) generates corrupted resources but `windres.exe` (in gcc/mingw) generates correct resources from the same input, or rather/at least, how I can work around it.

rpavlik:
A clue - we could try `#pragma code_page(65001)` to tell it the file is utf-8

PR #165, opened 2015-07-02
Version-awareness

rpavlik:
This is progress toward both overall self-knowledge of versions and solving #116 - code specifically producing those resource files isn't commited/pushed yet because it's acting bizarre, only compiling into proper version data in MinGW, not MSVC.


rpavlik:
@mars979 this is the branch that had the version stuff in it. Needs the UTF-8 pragma in the generated RC files, and presumably needs to be rebased.

PR #181, opened 2015-07-20
Gesture interface merge 

mars979:


inc/osvr/Common/RegisteredStringMap.h
```diff
@@ -0,0 +1,159 @@
...
+
+// Standard includes
+#include <string>
+#include <stdint.h>
+#include <vector>
+
+namespace osvr {
+namespace common {
+
+#define MAX_REGISTRY_SIZE 100
```
rpavlik: Nooo! Big code guideline violation. (And use an enlargeable structure instead of having a fixed size)

inc/osvr/Common/RegisteredStringMap.h
```diff
@@ -0,0 +1,159 @@
...
+
+    typedef char EntryName[100];
```
rpavlik: Nooo! Big code guideline violation - use std::string!

inc/osvr/Common/RegisteredStringMap.h
```diff
@@ -0,0 +1,159 @@
...
+
+    /// idea from the
+    /// http://www.ilikebigbits.com/blog/2014/5/6/type-safe-identifiers-in-c
+    template <class Tag, class impl> class ID {
```
rpavlik: I'd actually move this out into its own file as a typesafe handle, and use a policy/traits class for invalid values, etc.

inc/osvr/Common/RegisteredStringMap.h
```diff
@@ -0,0 +1,159 @@
...
+        bool empty() const { return m_val == 0xffffffff ? true : false; }
+
+        // this messes with implicit conversion (type casting)
+        /*
+        friend bool operator>(ID a, impl b) { return a.m_val > b; }
+        friend bool operator<(ID a, impl b) { return a.m_val < b; }
+        friend bool operator==(ID a, ID b) { return a.m_val == b.m_val; }
+        friend bool operator!=(ID a, ID b) { return a.m_val != b.m_val; }
+        */
+        impl m_val;
```
rpavlik: should be private.

inc/osvr/Common/RegisteredStringMap.h
```diff
@@ -0,0 +1,159 @@
...
+        ID() : m_val(0xffffffff) {}
+
+        // Explicit constructor:
+        explicit ID(impl val) : m_val(val) {}
+
+        // Explicit conversion to get back the impl:
+        // explicit operator impl() const { return m_val; }
+
+        // Implicit conversion to get back the impl
+        operator impl() const { return m_val; }
```
rpavlik: This shouldn't be here - should have a get() method or something otherwise we lose type safety.

inc/osvr/Common/RegisteredStringMap.h
```diff
@@ -0,0 +1,159 @@
...
+    typedef ID<struct StringTag, uint32_t> PeerStringID;
+
+    /// create a container for string registry
+    /// store the name and id
+    typedef struct StringRegistry {
+        char *name;
+        StringID id;
+    } StringRegistry;
+
+    typedef std::vector<StringRegistry> RegistryList;
```
rpavlik: Why not a map? This is a lot of data structures, are there tests?

inc/osvr/Common/SerializationTraits.h
```diff
@@ -510,6 +511,15 @@ namespace common {
                 f(val.data[2]);
             }
         };
+
+        template <>
+        struct SimpleStructSerialization<StringID>
+            : SimpleStructSerializationBase {
+            template <typename F, typename T> static void apply(F &f, T &val) {
+                f(val.m_val);
```
rpavlik: kind of looks like you're accessing a private member: I'd suggest having a member function named `value()` or similar that returns a reference.

inc/osvr/PluginKit/GestureInterfaceC.h
```diff
@@ -0,0 +1,82 @@
...
+@param dev Device token
+@param iface Gesture interface
+@param gestureName String name of gesture
+@param gestureState Current state of gesture (In process vs Completed)
+@param sensor Sensor number
+@param timestamp Timestamp correlating to Gesture data.
+*/
+OSVR_PLUGINKIT_EXPORT
+OSVR_ReturnCode
+osvrDeviceGestureReportData(OSVR_IN_PTR OSVR_GestureDeviceInterface iface,
```
rpavlik: The point in the gesture string map is that the user's code requests a gesture by name once, and holds on to that ID  using it instead of strings. If we use strings each time, we'd do nearly as well to just skip the string mapping. So, need another API method here to get a gesture ID. (API-wise, looks a bit like getting a message type ID)

inc/osvr/Util/ClientCallbackTypesC.h
```diff
@@ -200,6 +200,22 @@ typedef struct OSVR_EyeTrackerBlinkReport {
...
 
+/** @brief Type of string to identify gesture name */
+typedef char const *OSVR_GestureName;
+
+/** @brief Type of Gesture state */
+typedef uint8_t OSVR_GestureState;
+
+/** @brief Report type for a gesture event */
+typedef struct OSVR_GestureReport {
+    OSVR_GestureName gestureName;
```
rpavlik: Can we just return the local gesture ID here? Then we don't have issues with lifetime of strings, etc.

inc/osvr/Util/ClientReportTypesC.h
```diff
@@ -204,6 +204,45 @@ typedef struct OSVR_EyeTrackerBlinkReport {
...
+#define OSVR_GESTURE_DOUBLE_TAP "DoubleTap"
+#define OSVR_GESTURE_PINCH "Pinch"
+#define OSVR_GESTURE_FINGER_SPREAD "FingerSpread"
+#define OSVR_GESTURE_CIRCLE "Circle"
+#define OSVR_GESTURE_LONG_PRESS "LongPress"
+#define OSVR_GESTURE_OPEN_HAND "OpenHand"
+#define OSVR_GESTURE_CLOSED_HAND "ClosedHand"
+
+/** @brief Report type for a gesture event */
+typedef struct OSVR_GestureReport {
```
rpavlik: This is duplicated in `ClientCallbackTypesC.h`

src/osvr/Client/GestureRemoteFactory.cpp
```diff
@@ -0,0 +1,161 @@
...
+// Library/third-party includes
+// - none
+
+// Standard includes
+// - none
+
+namespace osvr {
+namespace client {
+
+    class NetworkGestureRemoteHandler : public RemoteHandler {
```
rpavlik: the word "network" here is not useful/needed

src/osvr/Client/GestureRemoteFactory.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+            auto gesture = common::GestureComponent::create();
+            m_dev->addComponent(gesture);
+
+            gesture->registerGestureHandler(
+                [&](common::GestureData const &data,
+                    util::time::TimeValue const &timestamp) {
+                    m_handleGesture(data, timestamp);
+                });
+            gesture->registerGestureMapHandler(
```
rpavlik: I'd prefer all the map handling to take place in the Component, to keep the logic centralized.

inc/osvr/PluginKit/GestureInterfaceC.h
```diff
@@ -0,0 +1,82 @@
...
+lifetime as the device token in order to send messages conforming to an
+Gesture interface.
+@param numSensors The number of sensors you will be reporting Gesture data :
+You can report 1+ sensors. This parameter may be subject to external limitations
+*/
+OSVR_PLUGINKIT_EXPORT
+OSVR_ReturnCode
+osvrDeviceGestureConfigure(OSVR_INOUT_PTR OSVR_DeviceInitOptions opts,
+                           OSVR_OUT_PTR OSVR_GestureDeviceInterface *iface,
+                           OSVR_IN OSVR_ChannelCount numSensors
```
rpavlik: Do we actually support multiple sensors for gesture per device?

src/osvr/Client/GestureRemoteFactory.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+// Standard includes
+// - none
+
+namespace osvr {
+namespace client {
+
+    class NetworkGestureRemoteHandler : public RemoteHandler {
+      public:
+        NetworkGestureRemoteHandler(vrpn_ConnectionPtr const &conn,
```
rpavlik: Is there a way to listen just for a single gesture's callbacks?

src/osvr/Client/GestureRemoteFactory.cpp
```diff
@@ -0,0 +1,161 @@
...
+            }
+            std::string gestureName = localGestureMap.getNameFromID(id);
+
+            if (gestureName.empty()) {
+                // could not find gesture name for this id, discarding report
+                return;
+            }
+            report.sensor = data.sensor;
+            report.state = data.gestureState;
+            report.gestureName = gestureName.c_str();
```
rpavlik: Can't do this - c_str() returns a temporary, so you can't expect it to hang around.

src/osvr/Client/PureClientContext.cpp
```diff
@@ -117,6 +120,7 @@ namespace client {
         ButtonRemoteFactory(m_vrpnConns).registerWith(m_factory);
         ImagingRemoteFactory(m_vrpnConns).registerWith(m_factory);
         EyeTrackerRemoteFactory(m_vrpnConns).registerWith(m_factory);
+        GestureRemoteFactory(m_vrpnConns).registerWith(m_factory);
```
rpavlik: This should be the only change committed in this file.

inc/osvr/Util/ClientReportTypesC.h
```diff
@@ -204,6 +204,45 @@ typedef struct OSVR_EyeTrackerBlinkReport {
...
+   process(ocurring)"
+    (should be used for continuous gestures, since discrete start and complete
+   at the same time) */
+#define OSVR_GESTURE_IN_PROCESS (1)
+
+/** @brief OSVR_GestureState value indicating "gesture is finished" */
+#define OSVR_GESTURE_COMPLETE (0)
+
+/** @brief a list of pre-set gestures available (can be expanded) */
+#define OSVR_GESTURE_SWIP_LEFT "SwipeLeft"
```
rpavlik: Is there a reason that these are spelled `SWIP` (no `E`)?

src/osvr/Common/GestureComponent.cpp
```diff
@@ -0,0 +1,236 @@
...
+        m_gestureNameMap.registerStringID(OSVR_GESTURE_DOUBLE_TAP);
+        m_gestureNameMap.registerStringID(OSVR_GESTURE_PINCH);
+        m_gestureNameMap.registerStringID(OSVR_GESTURE_FINGER_SPREAD);
+        m_gestureNameMap.registerStringID(OSVR_GESTURE_CIRCLE);
+        m_gestureNameMap.registerStringID(OSVR_GESTURE_LONG_PRESS);
+        m_gestureNameMap.registerStringID(OSVR_GESTURE_OPEN_HAND);
+        m_gestureNameMap.registerStringID(OSVR_GESTURE_CLOSED_HAND);
+    }
+
+    void GestureComponent::recordParent(Parent &dev) {
```
rpavlik: This is the purpose of having `m_parentSet()` - so you don't override this method if you want to hook in your behavior.

src/osvr/Common/GestureComponent.cpp
```diff
@@ -0,0 +1,236 @@
...
+            cb(data, timestamp);
+        }
+        return 0;
+    }
+
+    int VRPN_CALLBACK
+    GestureComponent::m_handleGestureMapRecord(void *userdata,
+                                               vrpn_HANDLERPARAM p) {
+        auto self = static_cast<GestureComponent *>(userdata);
+        auto bufwrap = ExternalBufferReadingWrapper<unsigned char>(
```
rpavlik: See the changes I made recently to clean this up from the eye tracking merge.

mars979: Fixed. I must have missed it because I changed the first message type but not the second one.

src/osvr/Common/GestureComponent.cpp
mars979: Fixed

inc/osvr/Util/ClientReportTypesC.h
mars979: Fixed the typos

src/osvr/Client/GestureRemoteFactory.cpp
mars979: Fixed

inc/osvr/PluginKit/GestureInterfaceC.h
mars979: That was my understanding that you can have gestures for both hands at least like in LeapMotion https://developer.leapmotion.com/documentation/cpp/devguide/Leap_Hand.html?proglang=cpp
And MYO armband allows multiple devices to be connected and interacted (so 1 armband on each hand is plausible)
https://developer.thalmic.com/docs/api_reference/platform/multiple-myos_8cpp-example.html
Looking at the spec now we defined a gesture not limited to just hands, so you can also have head or other part of the body perform a gesture, then using Kinect you can (in theory) have more than 2 sensors in a device.

rpavlik: OK, that's what I thought. Then we need to think about how best to expose the particular gestures in the path tree and let applications subscribe to just one - whether we just do that as a filter on the client end, or if we modify the path tree to support "sub-sensors" - something like `/plugin/device/gesture/0/pinch`

rpavlik: For now I'd say just a client library filter, have a way to register a callback optionally for just one gesture. The catch is that this means state is weird, since it will tell you the last gesture event to come through.

src/osvr/Client/GestureRemoteFactory.cpp
mars979: Not sure what you meant here. So the gesture component has a separate message that contains updated map of gesture names to IDs which client receives and "synchronizes" with its own.

inc/osvr/Util/ClientReportTypesC.h
mars979: Fixed, removed dups from `ClientCallbackTypesC.h`

inc/osvr/Common/RegisteredStringMap.h
mars979: Fixed to use map. Haven't made the tests, that's on todo list

inc/osvr/Common/RegisteredStringMap.h
mars979: Fixed

inc/osvr/Common/RegisteredStringMap.h
mars979: Fixed

inc/osvr/PluginKit/GestureInterfaceC.h
mars979: The name to ID map is kept inside the gesture component right now, which looks it up before sending to client. The gesture names get mapped to ID only once but you still need to look up ID by name every time you send it.

inc/osvr/Util/ClientCallbackTypesC.h
mars979: So right now the client doesn't have access to local gesture map to look up name. It is kept inside the gesture remote handler. Did you want to have the client apps be able to access the local map directly?

inc/osvr/PluginKit/GestureInterfaceC.h
rpavlik: OK, so that's going to be between somewhat inefficient and very inefficient.

mars979: Okay, so we would probably need another data structure in gesture component, like a hash table to provide fast access to IDs, maybe stl's unordered_map?

rpavlik: Unless you have a really good reason not to (type safety sadly not counting given that this is a C API), I'd say that the plugins and clients have access to the numeric IDs directly - that's how the rest of the system was designed and what I had foreseen.

zachkinstner:
Notes about Leap Motion gestures: https://github.com/OSVR/OSVR-Leap-Motion/issues/5

rpavlik:
The places where an Interface pointer is being held, should instead be using an `InternalInterfaceOwner`, which was just added to master recently.

mars979:
This pull request is ready to be merged

rpavlik:
OK - blocked on #218 - after that is finished (unit tests) and merged, this can be rebased on top of the new master. Most of all the "little" pieces that are independent have been pulled out and merged separately, so there will be ample use of `git mergetool` during that rebase, but it should result in a more digestable diff. I'd also suggest squashing related commits down into a single commit, especially early in the history where a lot of things changed since then.

mars979:
I complete unit tests merged #218 yesterday and rebased gesture branch on top of master. Also I fixed a few things that wouldn't let it build and some compiler errors.

rpavlik:
Will require fix to adjust for https://github.com/OSVR/OSVR-Core/commit/0a63fd5e091b4b615607f6ddde78acb6f2d49f2d once that branch is merged. Sorry...

rpavlik:
If this adds any additional typesafeids beyond those in stringids already in master, those will need to be adjusted for https://github.com/OSVR/OSVR-Core/commit/477c662c5ee093d4d139e5ffe18eabfa27f3fafb

PR #189, opened 2015-08-17
Matrix client api

rpavlik:
Pulls the JSON display descriptor parsing and projection/view matrix creation inside of Client/ClientKit, for simpler client code and improved flexibility in display description.


examples/clients/opengl-sdl/OpenGLSample.cpp
```diff
@@ -0,0 +1,290 @@
...
+            /// received
+            glMatrixMode(GL_MODELVIEW);
+            glLoadIdentity();
+            glMultMatrixd(viewMat);
+
+            /// For each display surface seen by the given eye of the given
+            /// viewer...
+            OSVR_SurfaceCount surfaces;
+            osvrClientGetNumSurfacesForViewerEye(disp, viewer, eye, &surfaces);
+            for (OSVR_SurfaceCount surface = 0; surface < surfaces; ++surface) {
```
russell-taylor: Someone has to call glClear() on the display.  Unfortunately, this has to happen once per display because it does not respect the glViewport.  I'm not seeing the call here, but it must be somewhere otherwise the cube would persist when you changed the viewing direction.  This makes it awkward to put the display loop inside the eye loop when there are multiple eyes on a given display.  This is not such a problem for rendertexture because each eye has its own buffer.

rpavlik: Well, I didn't write one...

So the question is then, whose problem is it to know what things are on the same GL display? In this case, since the app is creating the window, I'd say it's the app's problem (as in, don't need some API for "is this on the same GL display")

russell-taylor: In RenderManager, it is the problem of RenderManager because it set up the displays and such.  So, as I'm switching from render-to-screen to render-to-texture this is changing under the covers.  This is why we need all these callbacks to the app: start display, start eye, end eye, end display.  If the app needs to own this, it is hideously complicated and special-case and you get code drift.  That's why I'm trying to pull it into OSVR.  My target is an app that can write to different surfaces just based on the config file.  I think I'm getting there, because the example apps should not be changing as I swap out render-texture underneath them...

PR #190, opened 2015-08-18
OpenGL example with the distortion shader

rpavlik:
I stubbed this in, but don't actually have anything going (above the normal opengl sample) beyond GLEW linking and shaders being turned into header files.


rpavlik:
@russell-taylor @JeroMiya I'm assuming I can close and delete this given that RenderManager is open source now and this distortion method is effectively deprecated?

russell-taylor:
I think so.

PR #195, opened 2015-08-25
Container improvements

rpavlik:
These improve/add some util headers, needed for work on JointClientKit.


JeroMiya:
This looks good, though I'm surprised there isn't something like UniqueContainer in boost or the standard library.

rpavlik:
if there is, I haven't found it. You're right, that it's really simple, but I figure centralizing the code is better than re-writing it all over the place.

PR #196, opened 2015-08-27
Video based tracker new rebased for merge

russell-taylor:
When I rebased the project on master, I got a merge conflict in:
	tests/cplusplus/Util/CMakeLists.txt
	tests/cplusplus/Common/CmakeLists.txt

I resolved them both based on the non-head version of the files, which talked about factoring out things
	for gtest.

If that is okay, this should be ready to merge.  It mostly changes things in directories that are not yet in master, for video-based tracking.


CMakeLists.txt
```diff
@@ -173,6 +177,17 @@ if(BUILD_OPENCV_CAMERA_PLUGIN)
     list(APPEND OPENCV_MODULES_USED ${OPENCV_CAMERA_EXTRA_LIBS})
 endif()
 
+if(BUILD_VIDEOTRACKER_PLUGIN)
+    set(VIDEOTRACKER_EXTRA_LIBS
+        opencv_imgproc
+        opencv_features2d
+        opencv_calib3d
```
rpavlik: Do you know if these libs are the same in OpenCV 3? For instance, I know highgui's video capture got split out into `videoio` (see the lines above)

rpavlik:
I made some minor fixes, but I can't merge this for two reasons:

- license status of `qedit.h` - there's no license header, and if it originated in the Windows SDK it's probably not redistributable this way. Will have to work around it.
- Breaks non-MSVC builds (including mingw on Windows) and potentially MSVC builds where the set of headers installed is smaller than desired - unconditionally uses directshow without checking for its presence first.

rpavlik:
Other notes that I failed to submit before fixing them:

- `cv::SimpleBlobDetector` changed in OpenCV 3 - fixed.
- Need an include that defines `DIBSIZE` - err, or I just re-implemented it quick.

Also made one theoretical performance improvement (but should get from O(n^2) to O(n) in one portion of the algo!) and some smaller tweaks.

CMakeLists.txt
russell-taylor: I chased down this path briefly, but didn't get very far.  Judging from later commits, it looks like you got further down the path than I did.

russell-taylor:
The directx_camera_server.* and base_camera_server.* are from an open-source project at UNC.  

Good catch on the qedit.h file, which is what it says it is: copied from a random web location.  It used to be distributed with some of the Microsoft APIs, but apparently is not any more.  How do we work around this?  The file is basically an interface specification, which I think someone ruled is not copyrightable (I'm not a lawyer, and I don't recall the details of the case).

Regarding breaking the builds, that will require trying the compile on other architectures and using #defines to guard.  Looks like I'll chase the master on a branch and not merge for a while.

rpavlik:
OK, I rebased-out the bad qedit file, and instead am using CMake scripts to find the old platform sdk, which is still downloadable. Also adjusted it so we don't build the plugin on non-Windows at this point, since I didn't get a chance to figure out the clean slice point to cut out the directshow stuff. 

Because of the rebase, do not merge any previous work in or it would re-introduce the commit. Just start fresh from master (after I hit merge) and cherry-pick any commits missed.

russell-taylor:
Thank you for pulling this all together.  The CMake scripts don't seem to work on my laptop, at least not out of the box.  I wish we didn't need the DirectShow stuff, especially since OpenCV is using DirectShow internally.  Couldn't find the difference between what they are doing and what the video camera class is doing that made it work.  Any idea how I convince it to find DirectShow?

Thanks!
Russ

rpavlik:
Do you actually have one of those very old (6.0) windows SDKs installed? It will just surgically use the single file it needs. Pretty sure I put a link in the CMake message to the download.

russell-taylor:
Did not notice the message in the CMake output window.  Am now downloading and installing.  I'm impressed that you were able to find this and get it pulled in.

Basic install without example did not work, even though the qedit.h file is present.  Tried deleting the cache and restarting and no luck.

rpavlik:
The plugin is not installed by default. If you're having trouble building, will need more details

russell-taylor:
When I add the DirectShow directory to the CMake prefix variable, it finds it and builds it.  It runs against both stored files and the HDK, so I think we're good to delete all the branches and will commence doing so.

rpavlik:
Great. Could you tell me what directory you had to add to the prefix? I thought I had gotten all of them with the find module.

russell-taylor:
C:/Program Files/Microsoft SDKs/Windows/v6.0

PR #201, opened 2015-09-01
CMakeLists: Use the jsoncpp submodule as part of VRPN.

Armada651:
We've already checked out this submodule, so why not use it?


Armada651:
I noticed this PR currently only works when building the server, but not when you're only building the client. I'll have a closer look at it next week.

Armada651:
@rpavlik This PR works correctly now, but it depends on https://github.com/vrpn/vrpn/pull/109.

rpavlik:
I'd feel better just having the build done locally, inside OSVR's build system, but using the submodule, if the version in the submodule is sufficient - I haven't compared that version against the one I typically use recently. Doing so in the simplest way possible would leave some consumers of a pre-compiled library high and dry (or at least in need of building jsoncpp on their own), but there's no reason we couldn't support an external jsoncpp with the internal as fallback - the Linux distros whose packaging rules I'm familiar with would require this anyway (Debian really does not like vendored-in versions of libraries, and jsoncpp is widely used and widely packaged - and we're not terribly picky about the version we use).

Armada651:
> I'd feel better just having the build done locally, inside OSVR's build system, but using the submodule

Would a PR that implements that be acceptable?

Also, currently we depend on HID API being compiled in VRPN, the only reason I had to make jsoncpp a separate project is because it isn't compiled with VRPN if we're just building the client. So should the same be done for HID API at some point?

PR #203, opened 2015-09-03
Video based tracker improvements

russell-taylor:
This has the axes lined up the right way in the transforms, maintains a fast update rate, and has much less jitter.  The only changes should be to the video-based tracker plug-in.

### DIFF
--- a/plugins/videobasedtracker/BeaconBasedPoseEstimator.cpp
+++ b/plugins/videobasedtracker/BeaconBasedPoseEstimator.cpp
@@ -228,6 +228,29 @@ namespace vbtracker {
         }
 
         //==========================================================================
+        // Reproject the inliers into the image and make sure they are actually
+        // close to the expected location; otherwise, we have a bad pose.
+        if (inlierIndices.rows > 0) {
+          std::vector<cv::Point3f>  inlierObjectPoints;
+          std::vector<cv::Point2f> inlierImagePoints;
+          for (int i = 0; i < inlierIndices.rows; i++) {
+            inlierObjectPoints.push_back(objectPoints[i]);
+            inlierImagePoints.push_back(imagePoints[i]);
+          }
+          std::vector<cv::Point2f> reprojectedPoints;
+          cv::projectPoints(inlierObjectPoints, m_rvec, m_tvec, m_cameraMatrix,
+            m_distCoeffs, reprojectedPoints);
+          for (size_t i = 0; i < reprojectedPoints.size(); i++) {
+            if (reprojectedPoints[i].x - inlierImagePoints[i].x > 4) {
+              return false;
+            }
+            if (reprojectedPoints[i].y - inlierImagePoints[i].y > 4) {
+              return false;
+            }
+          }
+        }
+
+        //==========================================================================
         // Convert this into an OSVR representation of the transformation that
         // gives the pose of the HDK origin in the camera coordinate system,
         // switching units to meters and encoding the angle in a unit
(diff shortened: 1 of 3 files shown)


plugins/videobasedtracker/BeaconBasedPoseEstimator.cpp
```diff
@@ -212,10 +212,43 @@ namespace vbtracker {
         // do okay without using it, so leaving it out.
         // @todo Make number of iterations into a parameter.
         bool usePreviousGuess = false;
+        int maxIterations = 5;
+        cv::Mat inlierIndices;
         cv::solvePnPRansac(
             objectPoints, imagePoints, m_cameraMatrix, m_distCoeffs, m_rvec,
-            m_tvec, usePreviousGuess, 20, 8.0f,
-            static_cast<int>(objectPoints.size() - m_permittedOutliers));
+            m_tvec, usePreviousGuess, 5, 8.0f,
```
rpavlik: Based on your commit messages, I assumed the presence of a literal 5 here was a mis-resolved merge conflict, so when I rebased I used the variable name instead.

Would be good to have the 8.0f "magic number" named as well.

plugins/videobasedtracker/BeaconBasedPoseEstimator.cpp
```diff
@@ -212,10 +212,43 @@ namespace vbtracker {
...
+          std::vector<cv::Point2f> inlierImagePoints;
+          for (int i = 0; i < inlierIndices.rows; i++) {
+            inlierObjectPoints.push_back(objectPoints[i]);
+            inlierImagePoints.push_back(imagePoints[i]);
+          }
+          std::vector<cv::Point2f> reprojectedPoints;
+          cv::projectPoints(inlierObjectPoints, m_rvec, m_tvec, m_cameraMatrix,
+            m_distCoeffs, reprojectedPoints);
+          for (size_t i = 0; i < reprojectedPoints.size(); i++) {
+            if (reprojectedPoints[i].x - inlierImagePoints[i].x > 4) {
```
rpavlik: what's the literal 4 here?

rpavlik:
There's also an assert I keep hitting when building with OpenCV 3 (mingw64/msys2) - `OpenCV Error: Assertion failed (confidence > 0 && confidence < 1) in run, file C:/repo/mingw-w64-opencv/src/opencv-3.0.0/modules/calib3d/src/ptsetreg.cpp, line 178` - looks like the only thing in calib3d that's getting used is `cv::projectPoints` according to the DLL imports anyway.

rpavlik:
OK, have it running here after rebase, though the transform is goofed up.  Did quick profiling to see if there are any low-hanging fruit performance-wise: the plugin spends about 11.5% of its time in `osvr::vbtracker::OsvrHdkLedIdentifier::getId` and its child calls, which includes about 8.5% of total computation timedoing std::rotate, inclusive of internal implementation calls. Also looks like it's doing a bit of waiting on mutex related to osvr core, will dig into those later (esp. since if you count the video capture, it's spending way more time repeatedly sleeping in that thread...)

rpavlik:
Also, it looks like the SimpleBlobDetector includes a FindContours call, but that find contours might also be called separately - these are very expensive calls (inclusive of child calls, plugin spends 35% of its time in findContours), so if we can reduce duplication there it would be a big win.

rpavlik:
This is the orientation I'm getting. translations along both y and z are reversed, as well as rotations about those axes.

screenshot 091

plugins/videobasedtracker/BeaconBasedPoseEstimator.cpp
russell-taylor: Magic number I'm tuning to figure out where to cut off reprojections.  Will give it a name.

russell-taylor:
Added the performance improvements to my ToDo list. Thanks for running the traces!

russell-taylor:
Looking into the transforms now.

PR #207, opened 2015-09-10
Fix building for Linux using Boost 1.58.0

feilen:
There's an upstream bug in Boost that forgets to include this header and results in a compilation error. I believe this has been fixed for Boost 1.59.x, but in the mean time this will allow compilation without issues.

### DIFF
--- a/src/osvr/Common/NormalizeDeviceDescriptor.cpp
+++ b/src/osvr/Common/NormalizeDeviceDescriptor.cpp
@@ -24,6 +24,7 @@
 // limitations under the License.
 
 // Internal Includes
+#include <boost/type_traits/remove_cv.hpp>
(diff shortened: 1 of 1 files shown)


Armada651:
Nice, I was running into this problem before. Could you remove the merge commit and instead do a `git pull --rebase`?

feilen:
Sure, just a moment.

feilen:
Done! Sorry that took so long, I hadn't set up SSH keys for my desktop and
did all the work from SSH on my phone.

rpavlik:
Thanks for the fix!

Does this also fix the boost fusion-related issue mentioned in `BUILDING.md`?

feilen:
I don't see what you mean, building.md doesn't mention fusion. But it fixes
building for the issue in 1.58.0 that's fixed in 1.58.1.

Issue #208, opened 2015-09-10
JointClientKit API: custom plugin/display paths


JeroMiya:
The current search path list in the plugin host API searches paths that are relative to the executable API. However, this is not always the location where plugins are located. For instance, in an Android application, the executable path will be ```/system/bin/app_process32```, and the plugins (for JointClientKit) will need to be copied from the app bundle to its /data/data/com.packagename/files directory for it to have access to them.

This affects any search paths in the server or plugin APIs when using the JointClientKit - for instance, when explicitly specifying a display config file, or when auto-loading plugins or loading a specific named plugin via the JointClientKit APIs. There needs to be a way to customize the search paths.

You could just have one API call to override the root executable directory. This option might make it easier to catch all the places where we generate search paths without making a separate API for each search path (display json vs plugins, et. al.). The developer would only need to recreate the typical osvr_server file layout at the that directory. It might be easier to implement as well, requiring only that we centralize any code that looks for the "executable directory" to call an internal utility function that applies the override, if present.

Alternatively, you could have APIs for each search path set (plugins, files referenced in the server config json, etc...) to add one or more custom search paths for each separately, but this would be more complicated to implement and to use. It might be more flexible, however.

JeroMiya:
The referenced pull request adds the current working directory to the search path. Might be a good short term solution for Android. Doesn't solve the problem for files referenced from the config json.

PR #209, opened 2015-09-11
Added the current working directory to the plugin search path.

JeroMiya:
Specifically, the current working directory plus the OSVR_PLUGIN_DIR predefine subpath, and only if the CWD is not the same as the root (to avoid duplicate paths).

cc: https://github.com/OSVR/OSVR-Core/issues/208

### DIFF
--- a/src/osvr/PluginHost/SearchPath.cpp
+++ b/src/osvr/PluginHost/SearchPath.cpp
@@ -54,15 +54,15 @@ namespace pluginhost {
         // binDir now normalized to PREFIX/bin
         auto root = binDir.parent_path();
 
-        // current working directory
-        auto currentWorkingDirectory = boost::filesystem::current_path();
-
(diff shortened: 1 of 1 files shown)


rpavlik:
I think this looks OK to me - potential security stuff any time you change where we might run code from, but the code isn't being run with any special privilege so it should be OK.

JeroMiya:
Hmm.. good point on security. This might be an issue if they put osvr_server somewhere in the PATH and start it up from some random place in their file system. Should we put a "Don't do that" somewhere in the docs?

rpavlik:
Or, if this is needed for a particular platform, just ifdef it. There are other ifdefs in there as well already.

JeroMiya:
Good call. It now only adds the CWD on Android, and we'll punt for other platforms (not needed yet).

Issue #77, opened 2015-04-06
Be able to build an OpenCV-free version of ClientKit and deps


rpavlik:
If building just the client portion (for instance, in a submodule), we should be able to exclude OpenCV support on request to reduce dependencies. Change would require:

- Change the names of corresponding libraries (add `noimaging` to the end or something) so they didn't accidentally get used in a situation where imaging support was required
- Add aliases to exported config so that a version built with imaging would supply the noimaging target but not vice versa
- Add a configuration header to clientkit with a `HAVE_IMAGING` define inspected by at least the imaging header.
- Actually ifdef the corresponding code.

cc @Armada651

Armada651:
I'd really like to see this change soon, this issue currently blocks integration into the Dolphin Emulator. It would also be a boon to game developers that won't have to ship OpenCV DLLs that they're not even really using.

Armada651:
Ideally we'd get rid of the OpenCV dependency in the ClientKit completely, we can put the image data in a common data type and have the developer decide what kind of image processing library to use.

Armada651:
I see that OSVR [already provides such a common data type](https://github.com/OSVR/OSVR-Core/blob/master/inc/osvr/ClientKit/Imaging_decl.h#L63), which means that OpenCV is used for nothing more than a wrapper in the ClientKit. Wrapping the data should be the responsibility of the application and not the ClientKit.

rpavlik:
Internally, there is basically only the bits in Common/ImagingComponent that use OpenCV - I think there's a type constant header and their buffer-freeing/allocating functions to handle appropriately-aligned buffers to enable. Shouldn't be too hard to make this optional at the build-system level and churn out libraries (with names modified to indicate no imaging support) without that code. Happy to review or help where needed.

Armada651:
I don't really need OpenCV to be optional on the server-side, that would just complicate things. But if we can remove the dependency on the client-side as per PR #210 that would be perfect.

rpavlik:
merged!

Issue #53, opened 2015-03-09
PluginKit: don't make interface report methods require both the interface object and the device token


rpavlik:
We should store those together so you only need the interface object. This will break API and ABI, so the sooner the better.

godbyk:
I ported the button and tracker interfaces to use the `DeviceInterfaceBase` class (commit 2e24739e465d747654252339d691a3a32440de12).

godbyk:
Ported the imaging interface to use the `DeviceInterfaceBase` class (commit 1a4f17550386c8e40177dcd97ba68ed67b557db2).

Issue #44, opened 2015-03-09
Boost-free C++11 (or TR1?) version of the clientkit C++ bindings


rpavlik:
Per a presentation I read re: Ubisoft's dev. of games in C++ , they don't allow boost in their main tree, and I'm sure they're not the only ones. The main reason we're using Boost in the clientkit bindings is for relatively simple things (shared_ptr, etc) that we could get if the user agreed to a higher C++ standard (Not sure if TR1 suffices or if C++11 is needed, offhand)

(presentation is here: https://github.com/CppCon/CppCon2014/blob/master/Presentations/C%2B%2B%20in%20Huge%20AAA%20Games/C%2B%2B%20in%20Huge%20AAA%20Games%20-%20Nicolas%20Fleury%20-%20CppCon%202014.pdf I just read the slides, didn't watch the video)

rpavlik:
@yboger Can you get in touch with your game industry contacts to see if this is worthwhile?  We were going for minimum compiler requirements/max compatibility on the API wrapper (it works at least as far back as VS 2008), hence the usage of Boost instead of C++11 or TR1, but what I've seen so far of game engines is that they're mostly all already using VS 2013. What I really wouldn't want to see is the (slightly unsafe, less usable) C API being used (or a wrapper re-invented) in C++ games/engines because the wrapper uses things that aren't allowed.

yboger:
what's the downside of using Boost?

rpavlik:
Well, it's a huge library collection, and apparently, at least according to that presentation, it's banned in some codebases. I think such a ban would have one of these causes:

- to reduce deps (particularly for projects that basically stick every dependency in their source tree - for instance, see the massive 1.5gb repo of just the VR-related components of the Steam SDK)
- a mis-placed or over-strict philosophical stance: it's possible to write crazy-complicated meta-style impossible-to-maintain code or code that incurs unexpected performance hits with some Boost libraries, but some Boost libraries are so basic and useful they've ended up in newer versions of the C++ standard.

It's the latter category of libraries we use in the wrappers, hence why we could basically replace/alternate the boost usage with newer C++ standards. They are close enough that we could maintain essentially the same interface, just letting the user pick whether they want to require Boost or a newer C++ compiler than VS2008.

yboger:
so if we got rid of Boost, we'd need VS2010? Something newer?

rpavlik:
[...]

PR #210, opened 2015-09-12
ClientKit: Remove dependency on OpenCV from Imaging.

Armada651:
Wrapping the image data into an OpenCV type should be the responsibility of the application. OpenCV is a very large dependency and it should not be depended on just for a few lines of code.

Fixes issue #77.


inc/osvr/ClientKit/Imaging.h
```diff
@@ -94,29 +92,25 @@ namespace clientkit {
                                      const struct OSVR_ImagingReport *report) {
                 ImagingCallbackRegistration *self =
                     static_cast<ImagingCallbackRegistration *>(userdata);
-                ImagingReportOpenCV newReport;
+                ImagingReport newReport;
                 newReport.sensor = report->sensor;
                 newReport.buffer.reset(report->state.data,
                                        ImagingDeleter(self->m_ctx));
-                newReport.frame = cv::Mat(
```
rpavlik: This is just in the optional C++ header-only wrapper. While it's recommended, because of typing reasons, it's not strictly required for use of even the imaging interface. There's not much point in having this extra layer wrapping the callbacks without marshalling the data into the OpenCV type - that's what this file exists for. If you just want the bytes, there's the C interface.

inc/osvr/ClientKit/Imaging_decl.h
```diff
@@ -50,23 +49,19 @@ namespace clientkit {
...
     /// that the image buffer doesn't get freed before you're done using it.
-    struct ImagingReportOpenCV {
+    struct ImagingReport {
         /// @brief The device sensor number this frame came from.
         OSVR_ChannelCount sensor;
 
-        /// @brief An OpenCV Mat header for accessing the image data in the
-        /// buffer. Does not store the data internally (to reduce copies)!
-        cv::Mat frame;
-
```
rpavlik: This is the main piece that distinguishes the C++ wrapper from the C API. I suppose the lifetime management could be separated from the use of OpenCV, but either way the changes to any C++ `.h` files in this directory do not actually affect the build or runtime requirements of a client that does not explicitly use this file by choice.

src/osvr/Common/ImagingComponent.cpp
```diff
@@ -71,9 +67,7 @@ namespace common {
 
             template <typename T>
             void allocateBuffer(T &, size_t bytes, std::true_type const &) {
-                m_imgBuf.reset(reinterpret_cast<OSVR_ImageBufferElement *>(
-                                   cv::fastMalloc(bytes)),
-                               &cv::fastFree);
+                m_imgBuf.reset(reinterpret_cast<OSVR_ImageBufferElement *>(malloc(bytes)));
```
rpavlik: The reason that OpenCV was used here was to ensure that allocations would be appropriately aligned for optimized OpenCV operations that work on aligned buffers. For a non-OpenCV build of the core, these should at least be replaced with an aligned allocation - pretty sure I've got headers to do that in there.

rpavlik:
Thanks for working on this!  I'd really rather see a "halfway" approach - being able to build without OpenCV optionally, mangling the library names so it's clear imaging support might be degraded. (It's annoying that the C++ standard doesn't provide all the alignment guarantees needed to support efficient processing with SIMD instructions, GPU processing, etc, but them's the breaks.

None of the changes in the util headers or in the clientkit headers should be needed to build without OpenCV - you should be able to just turn off the building of the "header dependency tests" (which you'd probably want to do in a submodule situation anyway) and then, in proper C++ form, you won't pay for what you don't use.

Thanks for pointing me to this - GitHub's notification system is befuddling and I typically end up getting notified only about things I don't want to be...

inc/osvr/ClientKit/Imaging.h
Armada651: But I thought we didn't want to encourage people to downgrade to the C interface just to get rid of a dependency? See issue #44.

inc/osvr/ClientKit/Imaging_decl.h
Armada651: There is no lifetime management here, the `cv::Mat` explicitly specifies that it will not de-allocate the data buffer if it was initialized with a data pointer.

src/osvr/Common/ImagingComponent.cpp
Armada651: If you could point me to those aligned allocation functions I'll implement them here.

Armada651:
I'm not really convinced this is really useful, most people will use their own image processing library. They'll load the image into something like a texture and never even use OpenCV, but currently they'll have to pull it in as a huge dependency.

Even if we make it optional for the developers that do use OpenCV, this wrapper only saves them a few lines of code at most. Is it really worth the added complexity of having two different versions of this header and the ClientKit?

src/osvr/Common/ImagingComponent.cpp
Armada651: I will reimplement this part using `std::align`.

I'll also update the client example to show that having a dependency for OpenCV in the ClientKit is completely unnecessary even if the developer uses OpenCV.

inc/osvr/ClientKit/Imaging.h
Armada651: @rpavlik Just to quote the specific argument:

> What I really wouldn't want to see is the (slightly unsafe, less usable) C API being used (or a wrapper re-invented) in C++ games/engines because the wrapper uses things that aren't allowed.

inc/osvr/ClientKit/Imaging_decl.h
Armada651: @rpavlik Is there any lifetime management I'm not aware of? As far as I can see the `ImageBufferPtr` below is the class doing the actual lifetime management.

inc/osvr/ClientKit/Imaging.h
rpavlik: Touche - good point :D Probably could split this into two layers - a lifetime-management layer and then a opencv layer on top, so you can use whichever you like. Especially with Imaging at the moment,I really don't want people using the C API because you actually get a lock on an element of a shared-memory ring buffer, so failing to release that is a problem.

Armada651: Exactly, but don't we already have that opencv layer? `OpenCVTypeDispatch.h` is all a developer would need to wrap the buffer pointer into a matrix using the `cv::Mat` constructor, it would basically be a one-liner.

Armada651:
@rpavlik I'm done implementing those changes. I think benefits of removing the OpenCV dependency outweighs the benefits of having the wrapper and the lifetime management inside the same structure.

If we still want to take care of the lifetime management we should probably make a child class of `cv::Mat` that actually supports `std::shared_ptr`.

ManusMachina:
@rpavlik Any feedback on this PR?

Armada651:
@rpavlik Any feedback on this PR?

d235j:
I'm curious as to what (if anything) is holding this up, too; OpenCV is a fairly heavyweight dependency and while it's very useful, I don't think pulling in large dependencies is a great idea.

rpavlik:
I agree, it's primarily lack of time to review, as well as hesitancy regarding breaking API/ABI.

The ClientKit header patches, as long as they come with matching fixes to the corresponding clients, are fine and easy to approve: those are to semi-optional pieces of the API (the C++ header-only stuff - semi-optional just for imaging because not managing imaging lifetime correctly right now is the only way to freeze the server from the client), don't affect already-compiled client apps, but some extra work to update clients using that wrapper to the new API. In general, pretty good.

However, the change in Common (the change in allocation/deallocation in ImagingComponent) actually breaks our plugin ABI, and would result in all plugins that do imaging crashing the server.  (At least on Windows, mixing allocation and deallocation, or doing so over .dll boundaries, is a fast lane to crashtown).

Now, the core deallocating a buffer allocated by a plugin seemed like a good idea at the time because it meant the core didn't do a copy/allocation, and I didn't expect opposition to using OpenCV (considered it the de-facto standard and that delay-loading the DLL in common was good enough to hide the "cost"). However, in retrospect a function with a user-provided deleter would have been better (same advantages: reduced copying, clear ownership semantics, plus no mandatory dependency). I agree that this should be done (it's tripped up several folks), we just need to take it to the mailing list probably and hash out how to do it - probably safest and easiest to break the API at the same time removing the old function and creating a new, differently-named function to replace it that takes both the buffer and the deleter.  Haven't actually had a non-backward-compatible API/ABI break in pluginkit in months, so you see why I am hesitant :)  (There are other things I'd like to break too, see #53 which is already implemented internally, we're just ignoring one parameter - so one break invites the question of doing more breaks all at once or instead giving "version 2" functions in addition...)

CC some folks recently touching imaging stuff: @JeroMiya @zachkinstner

d235j:
@rpavlik are we at a point yet where OSVR is stable enough that breaking API changes are an issue? Obviously this would have to undergo more serious discussion (mailing list is not a bad idea here), but considering that OSVR is not yet widely used, it's better we do breaking API changes that are needed now rather than putting in place non-deal API. At some point it might be worthwhile doing a review of sorts going over the API and finalizing things.

I have to look at the allocation/deallocation stuff more closely. I've been through that mess on Windows myself and it's not something you want to deal with. It gets even worse when someone decides to replace operator new and delete, and it would not surprise me if some codebases do this.

Armada651:
@rpavlik The allocator and deallocator work exactly like their OpenCV counterpart. It stores the pointer to the original buffer at index -1. It should be no problem for it to deallocate buffers allocated by OpenCV and vice-versa. Have you tested whether it actually breaks the ABI?

rpavlik:
No, I haven't had a chance, but based on the number of crashes I've heard of when people didn't use the exact build of OpenCV we used to allocate their buffer, and the number of times cross DLL deallocation has bitten me, it makes me nervous. (Thinking of an ABI break as in change of behavior that affects library consumers, not directly the binary interface itself, of course. The subtle kind of break that's harder to test for) It would be great if it didn't break ABI, then we could put off the real fix for longer, but I simply haven't had a chance to test it on Windows. Easiest way would be to make a windows build then try to use a plugin (OpenCV camera, leap motion) compiled against a shipped binary snapshot. Testing on Android would be harder for poor @jeromiya

d235j:
This is likely to only be an issue on Windows due to the nature of static linking and (de)allocators on that platform. I've spent hours debugging issues caused by (de)allocations across DLL boundaries and this is something we should avoid if possible.
It does look like the changes here use raw malloc and free though, which should be safe.

Armada651:
@rpavlik I can do the test, but will that really be enough to convince you the changes are safe?

You're getting the crash because people are mixing the Visual C++ versions of the OpenCV DLLs. This change will actually fix the crash because OSVR will always do the allocation/deallocation with its own Visual C++ version.

rpavlik:
It's not that I need convincing the changes are safe, it's that if they happen to be ABI-safe right now that's purely coincidence, and the correct fix is to fix the API.

If you split this pull request into ClientKit-only and core, I think we can get stuff merged sooner.

Armada651:
Doesn't the ClientKit rely on `ImagingComponent`?

rpavlik:
Its implementation does, but its API doesn't. This accomplishes something (reducing required libraries for the client API) on its own.

Armada651:
I need to compile the implementation, that's the whole reason I'm making these changes.

So how do you want the API fixed? Will the plugin be responsible for allocation/deallocation or will the server be responsible?

rpavlik:
I think the plugin should be responsible as mentioned in https://github.com/OSVR/OSVR-Core/pull/210#issuecomment-146709467 but please email the devel list to have a public discussion so that other stakeholders can participate.

Armada651:
@rpavlik Could you send that email to the devel list? You know more about this than I do.

Armada651:
@rpavlik No response from the mailing list yet...

yboger:
@rpavlik I agree. Let's not wait any longer and just do it.

Armada651:
Okay, I'll make the allocation/deallocation API changes this week and then we can merge it.

Armada651:
@rpavlik I've tried looking at the code base today regarding memory management, but it is a lot more complicated than I anticipated. I don't think I'll be able to do that within a week and I honestly don't have time for it.

rpavlik:
I'll see when I can take a look at the pull request. If nothing else I can merge the API-only stuff easily. I've posted a followup on the mailing list (or at least I tried so, though I don't see it yet) laying out the ABI/API break stuff. Quite interesting that you were able to use a plugin from the snapshot (Windows, I'm assuming, since we don't publish other snapshot binaries?) with this PR - given how crash-happy that particular code has been, I'm pleasantly surprised.  We'll fix the API and whole thing soon here.

src/osvr/Common/ImagingComponent.cpp
```diff
@@ -71,9 +70,15 @@ namespace common {
 
             template <typename T>
             void allocateBuffer(T &, size_t bytes, std::true_type const &) {
-                m_imgBuf.reset(reinterpret_cast<OSVR_ImageBufferElement *>(
-                                   cv::fastMalloc(bytes)),
-                               &cv::fastFree);
+                // Allocate a memory pointer aligned to a boundary.
+                size_t space = bytes + sizeof(void*) + OSVR_ALIGN_SIZE;
+                void* buffer = malloc(space);
+                void* aligned = std::align(OSVR_ALIGN_SIZE, bytes, buffer, space);
```
rpavlik: I'm not terribly familiar with the C++11 alignment functionality yet - are we guaranteed here that `aligned > buffer` so that below, when we do `aligned[-1]` we're not writing out of bounds?

Armada651: Yeah that's an error, there should be a `buffer+1` there. Thanks for
catching it, will correct it tomorrow.

Armada651:
@rpavlik I've made the necessary changes, it's ready for review.

Armada651:
@rpavlik You mentioned that you'd like to move the allocator/deallocator to osvrUtil? Are there any other changes you need? I can make those changes this weekend so that the PR can be merged without any more work from your side.

rpavlik:
oh, thanks!  That would be great - yeah, just moved to Util, probably into an implementation file, with a public header exporting the functions, and maybe adjust the C++ wrapper on the pluginkit side so that it uses that, instead of the OpenCV matrix clone.
(The silly build stuff would take me longer to describe than to do.)

Armada651:
@rpavlik I'm done moving the allocator function into its own header. I'm not sure what to do with the PluginKit, using these new allocator functions there would result in copying a buffer which we want to avoid.

rpavlik:
There's already a buffer copy hidden there. I can take care of it - I put it there. I plan to review and merge this today.

rpavlik:
I've pulled this locally and am just making a few little changes before pushing it to master. Will leave this open until that's in and merged.

rpavlik:
OK, I've opened the new pull request #260 which is just this one, merged into master, with a few changes. Waiting on some help on making sure I didn't break code somebody else understands better than me, and a positive travis-ci result, then should be clear to go.

PR #212, opened 2015-09-14
New joint client kit test, with custom client options

JeroMiya:
Loads eye tracker plugin, instantiates of an interface, and registers a callback.


rpavlik:
Works nice - the joint client kit opengl sample works right now too.

PR #213, opened 2015-09-16
Added a C API sample for the imaging interface.

JeroMiya:
Do not merge yet. This sample exposes a possible bug in the C API for the imaging interface. It appears the osvrClientFreeImage calls are failing, possibly because the image objects are not being tracked when the callbacks are being called.

### DIFF
--- a/examples/clients/Imaging.c
+++ b/examples/clients/Imaging.c
@@ -1,5 +1,5 @@
 /** @file
-    @brief Implementation
+    @brief C Imaging interface sample
 
     @date 2015
 
@@ -8,6 +8,7 @@
     <http://sensics.com/osvr>
 */
 
+/*
 // Copyright 2015 Sensics, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


JeroMiya:
OK, this sample is working now, with the included fix to NetworkImagingRemoteHandler. Thanks @rpavlik for help in finding that bug. This pull request should be ready for review/merge.

examples/clients/Imaging.c
```diff
@@ -0,0 +1,90 @@
...
+        return -1;
+    }
+
+    OSVR_ClientInterface camera = NULL;
+    if(OSVR_RETURN_SUCCESS != osvrClientGetInterface(ctx, "/camera", &camera)) {
+        printf("Error, could not get the camera interface at /camera.\n");
+        return -1;
+    }
+
+    /// We keep a copy of the last report to avoid de-allocating the image
```
rpavlik: Can remove this boatanchor

examples/clients/Imaging.c
```diff
@@ -0,0 +1,90 @@
...
+    /// buffer until we have a new report.
+    //osvr::clientkit::ImagingReportOpenCV lastReport;
+
+    // Register the imaging callback.
```
rpavlik: I think we're still technically using just `/* */` for the C stuff in case it bothers old C compilers.

src/osvr/Client/ImagingRemoteFactory.cpp
```diff
@@ -88,8 +88,8 @@ namespace client {
             report.state.metadata = data.metadata;
             report.state.data = data.buffer.get();
             for (auto &iface : m_interfaces) {
-                iface->triggerCallbacks(timestamp, report);
                 iface->getContext().acquireObject(data.buffer);
+                iface->triggerCallbacks(timestamp, report);
```
rpavlik: Just occurred to me - the only case in which this would not be good is if somehow the list of interfaces was populated but there were no callbacks registered. We probably need to handle that case.

rpavlik: and need to acquire the object once for each callback that was registered...

PR #214, opened 2015-09-17
Additional api methods for getting display parameters

mars979:
Includes new methods for 
osvrClientGetNumDisplayInputs,
osvrClientGetDisplayDimensions,
osvrClientGetViewerEyeSurfaceDisplayInputIndex,
osvrClientGetViewerEyeSurfaceProjectionClippingPlanes,

### DIFF
--- a/inc/osvr/Client/DisplayConfig.h
+++ b/inc/osvr/Client/DisplayConfig.h
@@ -43,6 +43,12 @@
 namespace osvr {
 namespace client {
 
+    struct NoViewerEyeParam : std::runtime_error {
+        NoViewerEyeParam(std::string const &param)
+            : std::runtime_error("No viewer eye available for provided " +
+                                 param + " parameter!") {}
+    };
+
     namespace detail {
         typedef util::ContainerWrapper<
             std::vector<Viewer>, util::container_policies::const_iterators,
--- a/inc/osvr/Client/DisplayConfig.h
+++ b/inc/osvr/Client/DisplayConfig.h
@@ -80,6 +80,9 @@ namespace client {
                                              OSVR_SurfaceCount surface) const;
 
         OSVR_CLIENT_EXPORT bool isStartupComplete() const;
+        OSVR_CLIENT_EXPORT OSVR_DisplayInputCount getNumDisplayInputs() const;
+        OSVR_CLIENT_EXPORT const ViewerEye &
+        getViewerEye(OSVR_DisplayInputCount dispInputIdx) const;
 
       private:
         friend class DisplayConfigFactory;
@@ -129,6 +132,7 @@ namespace client {
         /// @todo right now only a single surface per viewer eye
         return getViewerEye(viewer, eye);
     }
+
 } // namespace client
 } // namespace osvr
 
--- a/inc/osvr/Client/DisplayConfig.h
+++ b/inc/osvr/Client/DisplayConfig.h
@@ -29,7 +29,9 @@
 #include <osvr/Client/Export.h>
 #include <osvr/Util/ClientOpaqueTypesC.h>
 #include <osvr/Util/ChannelCountC.h>
+#include <osvr/Client/DisplayInput.h>
 #include <osvr/Client/Viewer.h>
+#include <osvr/Client/Viewers.h>
 #include <osvr/Util/UniquePtr.h>
 #include <osvr/Client/InternalInterfaceOwner.h>
 #include <osvr/Util/ContainerWrapper.h>
@@ -43,18 +45,6 @@
 namespace osvr {
 namespace client {
 
-    struct NoViewerEyeParam : std::runtime_error {
-        NoViewerEyeParam(std::string const &param)
-            : std::runtime_error("No viewer eye available for provided " +
-                                 param + " parameter!") {}
-    };
-
-    namespace detail {
-        typedef util::ContainerWrapper<
-            std::vector<Viewer>, util::container_policies::const_iterators,
-            util::container_policies::subscript> ViewerContainerBase;
-    } // namespace detail
-
     class DisplayConfig;
     typedef unique_ptr<DisplayConfig> DisplayConfigPtr;
     class DisplayConfigFactory {
@@ -63,7 +53,7 @@ namespace client {
         create(OSVR_ClientContext ctx);
     };
 
-    class DisplayConfig : public detail::ViewerContainerBase {
+    class DisplayConfig {
       public:
         DisplayConfig(DisplayConfig const &) = delete;
         DisplayConfig &operator=(DisplayConfig const &) = delete;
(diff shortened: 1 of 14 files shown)


rpavlik:
OK, I rebased this to get out the wrong root I accidentally left in. I also ran clang-format at every commit and did some squash/fixup of commits (`git rebase -i`) - please run clang-format on stuff you touch before committing in the future, makes review and merging easier.

Dropped the last commit (adding an export declaration) because it's not needed for a method whose definition is inline with the class body.

inc/osvr/Client/DisplayConfig.h
```diff
@@ -80,6 +86,9 @@ namespace client {
                                              OSVR_SurfaceCount surface) const;
 
         OSVR_CLIENT_EXPORT bool isStartupComplete() const;
+        OSVR_CLIENT_EXPORT OSVR_DisplayInputCount getNumDisplayInputs() const;
+        OSVR_CLIENT_EXPORT const ViewerEye &
```
rpavlik: I really am not a big fan of this method - it implies there's only one viewer eye per display input, which is usually not the case. Display inputs don't know their surfaces, but surfaces know their display inputs.

inc/osvr/Client/ViewerEye.h
```diff
@@ -109,15 +110,23 @@ namespace client {
         getProjection(double near, double far,
                       OSVR_MatrixConventions flags) const;
 
+        /// @brief Gets clipping plane for a given eye
+        OSVR_CLIENT_EXPORT util::Rectd getRect(double near, double far) const;
```
rpavlik: Can leave off the two parameters and just pass `1.0` in the implementation for near, to get the clipping planes (plural) for the given *surface*.

inc/osvr/ClientKit/DisplayC.h
```diff
@@ -136,7 +173,8 @@ osvrClientCheckDisplayStartup(OSVR_DisplayConfig disp);
 
     @sa OSVR_ViewerCount
 
-    @return OSVR_RETURN_FAILURE if invalid parameters were passed, in which case
+    @return OSVR_RETURN_FAILURE if invalid parameters were passed, in
+   which case
```
rpavlik: This is the main negative side-effect of clang format: weird comment breaking. Easiest fix is to merge everything into a single line and let it break the whole thing at once. In this case, since this code wasn't touched, committing this change was an "oops" on my part presumably when I went through and formatted.

src/osvr/Client/DisplayConfig.cpp
```diff
@@ -112,6 +113,25 @@ namespace client {
                 params.k1.data[2] = k1.k1_blue;
                 distort = params;
             }
+
+            // get the number of display inputs
+            // if larger than one then we need to assign input index for each
+            // viewer eye
+            auto numDispInputs = desc.getNumDisplayInputs();
+            std::vector<uint8_t> displayInputIndices;
+            if (numDispInputs == 2) {
```
rpavlik: Is this really the right conditional? I think there's an enum for that purpose: something like FULL_SCREEN means each eye gets a video input to itself. Yes, this isn't fully orthogonal.

src/osvr/Client/DisplayConfig.cpp
```diff
@@ -112,6 +113,25 @@ namespace client {
...
+                displayInputIndices = {0, 1};
+            } else {
+                displayInputIndices = {0};
```
rpavlik: Could make this line `{0, 0}` and be able to drop a conditional down below.

src/osvr/Client/DisplayConfig.cpp
```diff
@@ -125,11 +145,22 @@ namespace client {
...
-                              desc.getPitchTilt().value(), distortEye));
+
+                OSVR_DisplayInputCount displayInputIdx;
+                // 2 eyes with 1 input
+                if (eyesDesc.size() > displayInputIndices.size()) {
+                    displayInputIdx = displayInputIndices.front();
+                }
+                // 2 eyes with 2 inputs OR 2 eye with 1 input
+                else {
+                    displayInputIdx = displayInputIndices[eye];
```
rpavlik: If you did `{0, 0}` above, you could always use this line, unconditionally.

src/osvr/Client/DisplayConfig.cpp
```diff
@@ -160,5 +191,38 @@ namespace client {
         }
         return true;
     }
+
+    OSVR_DisplayInputCount DisplayConfig::getNumDisplayInputs() const {
+
+        // assume there's 1 input but still check if there are more than one
+        // display inputs by checking every viewer eye (0 based index)
```
rpavlik: This is conceptually a little flawed - there should be an object called VideoInput or something like that, that has properties width and height, and that the display config owns a container of. Once we get more data on video inputs, we'll need that.

You're not assuming there is one input, so that bit of the comment can be removed.

src/osvr/ClientKit/DisplayC.cpp
```diff
@@ -130,10 +134,35 @@ OSVR_ReturnCode osvrClientFreeDisplay(OSVR_DisplayConfig disp) {
...
+OSVR_ReturnCode osvrClientGetDisplayDimensions(
+    OSVR_DisplayConfig disp, OSVR_DisplayInputCount displayInputIndex,
+    OSVR_DisplayDimension *width, OSVR_DisplayDimension *height) {
+
     OSVR_VALIDATE_DISPLAY_CONFIG;
-    return disp->cfg->isStartupComplete() ? OSVR_RETURN_SUCCESS : OSVR_RETURN_FAILURE;
+    OSVR_VALIDATE_DISPLAY_INPUT_ID;
+    OSVR_VALIDATE_OUTPUT_PTR(width, "width");
+    OSVR_VALIDATE_OUTPUT_PTR(height, "height");
+    auto viewport =
```
rpavlik: This is definitely wrong. See previous comment about display model - should have a DisplayInput object with these properties.

rpavlik:
OK, review done - awaiting revision

inc/osvr/Client/DisplayInput.h
```diff
@@ -0,0 +1,63 @@
...
+
+// Standard includes
+// - none
+
+namespace osvr {
+namespace client {
+    class DisplayConfigFactory;
+    class DisplayInput {
+      public:
+        DisplayInput(DisplayInput const &) = delete;
```
rpavlik: For now, you can remove this line (deleting copy constructor), the deleting-copy-assignment, and the manual implementation of a move-constructor - nothing here is noncopyable.

inc/osvr/Client/ViewerEye.h
```diff
@@ -109,15 +110,23 @@ namespace client {
...
+
       private:
         friend class DisplayConfigFactory;
         ViewerEye(
             OSVR_ClientContext ctx, Eigen::Vector3d const &offset,
             const char path[], Viewport &&viewport, util::Rectd &&unitBounds,
             bool rot180, double pitchTilt,
-            boost::optional<OSVR_RadialDistortionParameters> radDistortParams);
+            boost::optional<OSVR_RadialDistortionParameters> radDistortParams,
+            OSVR_DisplayInputCount displayInputIdx);
```
rpavlik: sorry I will have caused a merge conflict here. I'll do the merge myself so you don't have to rebase.

inc/osvr/ClientKit/DisplayC.h
mars979: Fixed

inc/osvr/Client/DisplayConfig.h
mars979: Fixed with a separate DisplayInput object

inc/osvr/Client/ViewerEye.h
mars979: Fixed

src/osvr/Client/DisplayConfig.cpp
mars979: Fixed

src/osvr/Client/DisplayConfig.cpp
mars979: You're right, fixed.

src/osvr/Client/DisplayConfig.cpp
mars979: Fixed

src/osvr/Client/DisplayConfig.cpp
mars979: Fixed with a separate DisplayInput object

inc/osvr/Client/DisplayInput.h
mars979: Ok, removed those.

mars979:
I pushed the changes based on the review comments above.

PR #217, opened 2015-09-21
String buffer builder

rpavlik:
A better way to handle client APIs that return strings.


rpavlik:
cc @Armada651  and @feilen since this removes a little bit of Boost.

feilen:
That'll remove some of the ugliness in my current branch, I'll pull that and remove my related changes later tonight.

PR #220, opened 2015-09-24
Refactoring in pursuit of analysis plugin API

rpavlik:
This is some internal refactoring, primarily work that simplifies the existing PureClientContext (used in ClientKit) and JointClientContext (used in JointClientKit) by pulling out functionality into a more single-purpose class.


src/osvr/Client/TrackerRemoteFactory.cpp
```diff
@@ -142,21 +142,12 @@ namespace client {
...
-            opts.reportOrientation = true;
-        } else {
-            /// pose and tracker both imply full reports (?)
-            opts.reportOrientation = true;
-            opts.reportPosition = true;
-            opts.reportPose = true;
-        }
+        opts.reportOrientation = true;
+        opts.reportPosition = true;
+        opts.reportPose = true;
```
mars979: Is setting all options to true for Tracker reports temporary solution until you add code to set from descriptor?

rpavlik: Yes. This was the only code path that was ever actually getting called anyway, which is why it was confusing. There was no way to get into any other path but the `tracker` one without making a really weird device descriptor.

PR #222, opened 2015-10-01
Add default constructor to DisplayInput to fix compile with Clang and libc++

d235j:
The lack of a default constructor for the DisplayInput class causes a compile error at https://github.com/OSVR/OSVR-Core/blob/master/src/osvr/Client/DisplayConfig.cpp#L150 . This is because resize( size_type count ) requires a default constructor, in case the vector is being enlarged. Obviously the vector is not being enlarged in this instance, but this does trip up Clang when using libc++.

### DIFF
--- a/inc/osvr/Client/DisplayInput.h
+++ b/inc/osvr/Client/DisplayInput.h
@@ -40,6 +40,7 @@ namespace client {
     class DisplayConfigFactory;
     class DisplayInput {
       public:
+        DisplayInput();
         DisplayInput(OSVR_DisplayDimension width, OSVR_DisplayDimension height);
 
         OSVR_CLIENT_EXPORT OSVR_DisplayDimension getDisplayWidth() const;
--- a/src/osvr/Client/DisplayInput.cpp
+++ b/src/osvr/Client/DisplayInput.cpp
@@ -33,6 +33,9 @@
 
 namespace osvr {
 namespace client {
+    DisplayInput::DisplayInput() {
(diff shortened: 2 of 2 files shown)


rpavlik:
Wow, thanks for being so quick on the pull request! I actually saw the error on the CI last night and fixed it this morning before I saw your pull, but thanks for contributing to OSVR! Please do stick around and contribute again - and if you're not already on the dev list, please join: http://lists.getosvr.org/listinfo.cgi/osvr-devel-getosvr.org

d235j:
Sure, thanks! I'll join the list.
I did send you a direct email regarding Mac OS X support; did you receive it?

rpavlik:
Ah, good to "meet" you! I did receive that, thank you very much! Forwarded it to my "work" email last night and was planning on replying today - again, you beat me to it :)  Very happy to see you got it working. Would be happy to have that re-posted on the mailing list. No need to wait for the full thing to send pull requests, bits and pieces are fine, and no need to actually test with an HDK, rift, or other supported hardware yet - there are lots in the community who'd be more than happy to try :)

rpavlik:
(and I'm glad we, mainly I, didn't break too much stuff on the Mac in the meantime...)

rpavlik:
Can you see if the latest master still needs this patch to build? I think my fix should have solved the problem in a different way, but wanted to double check before I closed the pull.

d235j:
It builds fine — we can close this.

PR #223, opened 2015-10-01
Mac fixes, part one

d235j:
This is the first round of fixes; I have not completed the implementation for USBSerialDevInfo yet. There will be another PR for the CMake SDL2 changes, though that's rather tricky as I will need to test on Linux and MinGW to avoid breakage.


rpavlik:
Looks great, thanks!  You can just temporarily disable detecting SDL2 on Mac if you want, since that's just for some examples.

Issue #193, opened 2015-08-20
OSVR-Core compile fails with Boost 1.59.00


russell-taylor:
Original complaint seems to be about line 33 in convert.hpp: error C2903: 'apply' : symbol is neither a class template nor a function template (F:\taylorr\Personal\Work\consulting\Sensics\OSVR\src\OSVR-Core\src\osvr\Common\ClientInterface.cpp)

8>  ClientContext.cpp
8>  ClientInterface.cpp
8>  IPCRingBuffer.cpp
9>------ Build started: Project: header_cpp_39, Configuration: RelWithDebInfo x64 ------
10>------ Build started: Project: header_cpp_57, Configuration: RelWithDebInfo x64 ------
11>------ Build started: Project: header_cpp_59, Configuration: RelWithDebInfo x64 ------
9>  main39.cpp
10>  main57.cpp
11>  main59.cpp
8>C:\local\boost_1_59_0\boost/fusion/container/map/detail/cpp03/convert.hpp(33): error C2903: 'apply' : symbol is neither a class template nor a function template (F:\taylorr\Personal\Work\consulting\Sensics\OSVR\src\OSVR-Core\src\osvr\Common\ClientInterface.cpp)
[...]

PR #225, opened 2015-10-03
Support build with Boost 1.59.0

d235j:
I reviewed the changelog at http://www.boost.org/users/history/version_1_59_0.html and I see no changes in Interprocess that should affect us. There are changes to Fusion which appear to cause issues on Windows though, as we can see in https://github.com/OSVR/OSVR-Core/issues/193 — I haven't looked at these yet.

### DIFF
--- a/src/osvr/Common/IPCRingBuffer.cpp
+++ b/src/osvr/Common/IPCRingBuffer.cpp
@@ -51,7 +51,7 @@ namespace common {
 
 /// Some tests that can be automated for ensuring validity of the ABI level
 /// number.
-#if (BOOST_VERSION > 105800)
+#if (BOOST_VERSION > 105900)
 #error                                                                         \
     "Using an untested Boost version - inspect the Boost Interprocess release notes/changelog to see if any ABI breaks affect us."
 #endif


d235j:
As a note, I had all sorts of issues with C2719 in Boost Intrusive (used by Interprocess) when attempting to compile on Windows 32-bit with VS2015. The specific error looked like "formal parameter with forced alignment won't be aligned" or so.

rpavlik:
Thanks for looking into this. Happy to merge it. I've got a replacement for the fusion usage almost ready to merge, but that interprocess issue sounds a pain, please open a separate issue for it.

d235j:
At this point, Fusion works so I don't know if replacing it is even necessary (though changing the location of the #define and adding a comment indicating why that #define is there might be a good idea). I'd still add this as a boost-free code path if this is something that will affect ClientKit and the API — if not, it probably doesn't really matter.

rpavlik:
OK, thanks. VS2013 will be easier to get building everywhere, since that's what we're using for production.

Fusion has been the biggest trouble spot for compiling with different versions, and it's really kind of overkill for what we're using (type-keyed data structure - which can be done pretty easily with variadic templates and tuple), hence the work to replace it.

PR #226, opened 2015-10-04
Improve rpath handling

d235j:
This improves rpath handling on both Mac OS X and Linux so that the rpath and library paths are correct both in the source tree and after installation. The binaries are set up to look in ../lib/ for the libraries, unless the install prefix is a system directory (specified by CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES), in which case this is left up to the system to decide.

I have not tested this change on Windows. This brings things to a point where I can easily use Travis for both Mac and Linux.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,10 @@ if(ANDROID AND CMAKE_VERSION VERSION_LESS 3.1.0)
     message(FATAL_ERROR "You must use CMake 3.1.0 or greater to build for Android.")
 endif()
 
+if (NOT DEFINED CMAKE_MACOSX_RPATH)
+    set(CMAKE_MACOSX_RPATH ON)
+endif()
+
 set(LOCAL_CMAKE_MODULE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/cmake-local")
 # Custom CMake modules from https://github.com/rpavlik/cmake-modules
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake" "${LOCAL_CMAKE_MODULE_DIR}")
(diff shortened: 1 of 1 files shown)


rpavlik:
Looks pretty good to me. The only catch is that the install prefix at configure time isn't necessarily where a package gets installed to (consider cpack as an example), so it would be preferable if this just set the default of a GUI cache variable, vs setting it directly (or some other way of permitting a user to override)

rpavlik:
Oh, and nothing rpath related affects Windows builds, so no worries there ☺

d235j:
I'll look into that — it should be easy enough to make this overridable with a cmake flag.

d235j:
The documentation at https://cmake.org/Wiki/CMake:Packaging_With_CPack indicates that CPack uses the install target generated by CMake internally — which means it should be setting the rpath correctly on package generation. If it is doing this, the logic in here should work properly. I'd like to test this though...

rpavlik:
I think the catch I'm concerned about is the fact that it's usually "not best practices" in CMake to refer to CMAKE_INSTALL_PREFIX in your build script, since that can be overridden at install time (`make install` does something like `cmake -P cmake_install.cmake`, and you can do `cmake -DCMAKE_INSTALL_PREFIX=yourdest -P cmake_install.cmake` to get the equivalent of something similar to `make install DESTDIR=yourdest` - commands are all from memory, but I used something like that with stow, and don't know if homebrew uses it too - or if it even matters, since the purpose in those cases is to compile-in an install directory different from where you're initially putting it but where it will be able to find itself later.)

That said, this is nitpicking and probably not a good reason to not merge. If you're fine with it, I'll merge this and your travis scripts, then just open an issue for overriding this behavior. Thanks so much for doing this! I'll wait to hear back from you before merging.

d235j:
Hmm, looking more closely, you're still going to run into the fact that the rpaths for libfunctionality and boost are not relative... I'll have to look deeper into this. If you'd like to, feel free to merge, open a new issue, and assign it to me.

JeroMiya:
Android build looks good with this PR.

PR #227, opened 2015-10-04
Add Travis support for building

d235j:
Yes, the script is ugly, but it works for both Mac and Linux. This is probably not final — I'd like to add testing to it once I have this working (see https://travis-ci.org/d235j/OSVR-Core).

Please review https://github.com/OSVR/OSVR-Core/pull/226 first — some functionality here depends on that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+language: cpp
+compiler:
+  - clang
+os:
+  - osx
+before_install:
+  - brew update
+  - brew tap homebrew/science
+  - brew tap d235j/osvr
+install:
+  - brew uninstall json-c
+  - brew -v install libusb
+  - brew -v install jsoncpp --HEAD
+  - brew -v install libfunctionality --HEAD
+  - brew -v install opencv
+script:
+  - git submodule update --init --recursive
+  - mkdir build
+  - cd build
+  - 'cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_OSX_ARCHITECTURES=x86_64'
+  - make all -j2
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,19 +3,64 @@ compiler:
   - clang
 os:
   - osx
+  - linux
 before_install:
-  - brew update
-  - brew tap homebrew/science
-  - brew tap d235j/osvr
+  - "if [ $TRAVIS_OS_NAME = 'osx' ]; then brew update; fi"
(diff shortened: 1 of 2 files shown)


rpavlik:
It looks like the latest build labeled as "Mac" actually built on Linux - does that make sense? https://travis-ci.org/d235j/OSVR-Core/jobs/83596775

d235j:
32.1 built on Mac and 32.2 built on Linux. See https://travis-ci.org/d235j/OSVR-Core/builds/83596772.

rpavlik:
Ah, silly me. Must have been rather tired to confuse those icons.

rpavlik:
(Any reason why you're forcing use of gcc on linux when the env. seems to suggest clang? We should be able to build with both...)

d235j:
Oh, I meant to use gcc — the environment is forcing clang because I forgot to add compiler: gcc to the .travis.yml, and because I'm using a newer compiler than the one included with Ubuntu 12.04. Should I be using the included compiler, which doesn't have C++11 support?

rpavlik:
GCC in precise doesn't have C++11 support? (or maybe Clang in precise? That I could believe, always used a ppa when I was still using Ubuntu) Use whatever you like, I was just curious

d235j:
They provide GCC 4.6, which does not support the --std=c++11 flag (but supports the older --std=c++0x flag), and lacks some language features (which I'm not sure we're using).

If you think we should be compiling with older compilers, or older boost, I can modify the Travis config for that.

rpavlik:
Well, our (other, Jenkins) CI builds on an (updated) debian wheezy using:

- gcc 4.7.2 (using the c++0x thing)
and
- clang 3.5.0

both with Boost 1.55 (just backported the package from a newer debian release), in addition to the released binaries (build with VS2013, Boost 1.57)

Locally, I build with the same environment as the CI for Windows (2013/1.57), as well as a MSYS/MinGW64 build (GCC 5.2, boost 1.57, ninja), and my Linux VM is Debian Jessie (whatever gcc and boost it comes with, and the latest clang from their deb repo - it's not powered up right now, can get details if needed)

d235j:
Cool. Don't forget to log into Travis and turn it on for this repo.

rpavlik:
Yep, just did that and pushed a tiny update to see if it triggers.

rpavlik:
thank you very much for working on this!

PR #229, opened 2015-10-04
Properly find SDL2 framework on Mac

d235j:
This replaces the FindSDL2.cmake script with a slightly-modified one from the Dolphin project (which is nearly identical to the one at https://github.com/tcbrindle/sdl2-cmake-scripts). This script properly locates the SDL2.framework that is often used on Mac OS X instead of the libraries. (It also locates the libraries, if that's what one is using and if the framework is not present.)

I have not tested this on Windows! I'm afraid this may cause breakage there. This works correctly on Linux.

### DIFF
--- a/cmake/FindSDL2.cmake
+++ b/cmake/FindSDL2.cmake
@@ -1,179 +1,188 @@
-# - Find SDL2
-# Find the SDL2 headers and libraries
+# Locate SDL2 library
+# This module defines
+# SDL2_LIBRARY, the name of the library to link against
+# SDL2_FOUND, if false, do not try to link to SDL2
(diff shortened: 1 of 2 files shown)


rpavlik:
I'm not sure the license of the stuff from dolphin, but I'll see if I can use this information to update the one I wrote for Mac support. I'll cc you in a pull request there to test it out.

d235j:
@rpavlik that particular file is specified as BSD-licensed in the header comments, so I figured it would not be an issue (and I asked someone from Dolphin to confirm).

d235j:
Closing in favor of https://github.com/OSVR/OSVR-Core/issues/236.

Issue #177, opened 2015-07-10
boost::fusion-related compile issue


rpavlik:
After merging the EyeTracker branch, we started getting a build error with Boost 1.49.0 as included with Debian Wheezy. Specifically, here's what we observed:

64d97d4ef58a0faf0d3ac3899c7c032dfb875d17 - builds fine with 1.49 (wheezy), 1.55 (jessie), and 1.57 (windows and android)
1b535b21bc68e247ec38ac7327460ac7e808dc79 - Builds fine with 1.55 (jessie), and 1.57 (windows and android), but does not build properly with 1.49 (wheezy)

The following commits up through and including d5b68a9c8f0b70416637f5b610ea114479b36c68 were (increasingly desperate) attempts to work around the issue. Some might have been useful, but none fixed the builds. The result of building that last revision include this log:

```
[ 28%] Building CXX object src/osvr/Common/CMakeFiles/osvrCommon.dir/ClientContext.cpp.o
In file included from /var/lib/jenkins/jobs/OSVR-Core-Build-Linux-Matrix/workspace/COMPILER/clang/label/Linux/src/osvr/Common/ClientContext.cpp:27:
In file included from /var/lib/jenkins/jobs/OSVR-Core-Build-Linux-Matrix/workspace/COMPILER/clang/label/Linux/inc/osvr/Common/ClientInterface.h:32:
In file included from /var/lib/jenkins/jobs/OSVR-Core-Build-Linux-Matrix/workspace/COMPILER/clang/label/Linux/inc/osvr/Common/InterfaceState.h:29:
In file included from /var/lib/jenkins/jobs/OSVR-Core-Build-Linux-Matrix/workspace/COMPILER/clang/label/Linux/inc/osvr/Common/ReportMap.h:39:
In file included from /usr/include/boost/fusion/include/map.hpp:10:
In file included from /usr/include/boost/fusion/container/map.hpp:13:
/usr/include/boost/fusion/container/map/convert.hpp:24:30: error: implicit instantiation of undefined template 'boost::fusion::detail::as_map<11>'
            typedef typename gen::
                             ^
[...]

PR #230, opened 2015-10-04
Fix fusion vector size

d235j:
This fixes https://github.com/OSVR/OSVR-Core/issues/193 (and probably https://github.com/OSVR/OSVR-Core/issues/177 too as the error output looks very similar, but I haven't tested). I realized that this was going on by looking carefully at the "notes" that MSVC produced and noticing that the pairs at the beginning of the generated list were getting trashed. This probably affects other platforms too, but is just not caught...

(This is partially documented here: http://www.boost.org/doc/libs/1_59_0/libs/fusion/doc/html/fusion/container/vector.html — what is not documented is that mpl vector to fusion vector conversions are affected.)

### DIFF
--- a/inc/osvr/Common/ReportTypes.h
+++ b/inc/osvr/Common/ReportTypes.h
@@ -26,6 +26,7 @@
 #define INCLUDED_ReportTypes_h_GUID_BB8F47AD_5F82_439E_C3F9_7C57D016D3C9
 
 // Internal Includes
+#define FUSION_MAX_VECTOR_SIZE 20 //Default is 10, too small for typelist
 #include <osvr/Util/ClientCallbackTypesC.h>
 
 // Library/third-party includes


d235j:
Hmm I'm not sure if that should go here or rather in https://github.com/d235j/OSVR-Core/blob/mpl-vector-size/inc/osvr/Common/ReportMap.h — the issue seems to be that the mpl vector gets transformed into a fusion vector later on...

d235j:
Yes, this does indeed fix #177. Looking at the changes in https://github.com/OSVR/OSVR-Core/commit/1b535b21bc68e247ec38ac7327460ac7e808dc79, the vector size did increase above 10.

rpavlik:
Thank you for tracking this down! I really do not enjoy debugging Boost Fusion...

d235j:
Any thoughts on my comments above?

rpavlik:
If you'd like to move it around and if it makes more sense to you that way, please feel free. Like I said, the Fusion usage is hopefully not long for this world, since we can rely on having C++11 variadic templates and tuple.

rpavlik:
(see https://github.com/OSVR/OSVR-Core/tree/tp-test and https://github.com/OSVR/OSVR-Core/tree/typepack-improvements for details - basically getting things compiling properly with VS2013 is the blocker)

PR #231, opened 2015-10-04
Remove unnecessary SourceAnnotations.h include

d235j:
&lt;CodeAnalysis/SourceAnnotations.h&gt; is automatically included from &lt;sal.h&gt; when necessary, and including it here separately breaks build for the C-only components. This is necessary for clean building with VS2015.


rpavlik:
Ah, didn't realize that, thought it was required to get the new style sal annotations. Thanks also for investigating visual studio 2015 - I hadn't yet investigated updating the build environment, but as it's closer to actual c++11 it should make building easier.

PR #232, opened 2015-10-06
Adjust FindSDL2 to hopefully be able to find the framework on Mac.

rpavlik:
cc @d235j - hopefully you can test this. It should follow/respect the preferences set in CMake regarding whether to find frameworks first, last, never, and should also correctly handle linking to cocoa.

### DIFF
--- a/cmake/FindSDL2.cmake
+++ b/cmake/FindSDL2.cmake
@@ -56,15 +56,6 @@ endif()
 
 include(FindPackageHandleStandardArgs)
 
-find_path(SDL2_INCLUDE_DIR
-	NAMES
-	SDL_haptic.h # this file was introduced with SDL2
-	HINTS
-	${SDL2_INCLUDE_HINTS}
-	PATHS
-	${SDL2_ROOT_DIR}
-	ENV SDL2DIR
-	PATH_SUFFIXES include include/sdl2)
 find_library(SDL2_LIBRARY
 	NAMES
 	SDL2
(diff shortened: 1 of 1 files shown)


d235j:
It seems like your regex is broken. Additionally, it's not properly generating the -framework Cocoa and -framework SDL compiler/linker flags necessary to build against a framework on OS X, but is rather treating the framework as a plain old .a library. (If a framework is located somewhere other than in a system path, it is additionally necessary to add -F/path/to/dir/containing/framework. Linking against a framework on OS X cannot be done by -l or the same way that you would link a library statically for various reasons.)

Any specific reasons not to use the CMake-derived FindSDL2? It's more flexible, and feels a bit cleaner to me, and properly links the frameworks. At the very least, though, $SDL2DIR should be supported to provide the path to the prefix where the SDL2 libs are installed.

d235j:
By the way, a bit more research indicates that FindGLUT.cmake is very likely broken on newer versions of OS X. You used to be able to link a framework by putting its full path on the compiler command line, but that changed sometime recently and the framework instead can contain a tbd file which is a "text-based stub library" that tells the compiler where the actual binaries can be found. Therefore the framework binary no longer even exists at the guessed full path!

rpavlik:
RE: the regex: doesn't surprise me, my regexes are usually wrong the first time, especially when I don't have a way to test them.

Apparently then this report is out of date. http://public.kitware.com/Bug/view.php?id=14105

I'd  prefer to stick with the self-developed file because
- it's more modern style (using imported targets) and can thus handle things like compile definitions as a part of the interface (for instance SDLmain's stuff)
- Spent a lot of time getting it to work untouched (at most adding something to CMAKE_PREFIX_PATH or the root dir variable) with the upstream binaries for MSVC and MinGW (the latter of which are somewhat weird) as well as the SDL2 package included with MSYS2/MinGW64

Hopefully we can figure out how to properly create an imported target on recent Mac OS X for a framework - sounds like that framework change is a nice wrench in the works for anyone who dares use a non-XCode-exclusive build..

In the meantime, as long as the existing script doesn't actually break the configure process, all that not finding SDL2 is causing is disabling a few small OpenGL samples from the build, which I think is non-critical. It's not actually breaking the mac build, right? (In the travis log I saw, it wasn't finding SDL2 anyway)

d235j:
Yeah, that report is out of date. It does look like Apple's breaking changes are very recent — possibly appearing in Xcode 7 and OS 10.11, which was released as a beta in June and has just hit general availability a week ago.

I don't have a problem with the existing script, as long as is responds to SDL2DIR which it seems like it does.

As for SDL2Main, it seems their approach is to use the SDL2_BUILDING_LIBRARY flag for things that don't need SDL2Main...

Regarding frameworks, I was considering potentially building OSVR as a framework, but that is a discussion that doesn't really belong here at this time.

d235j:
By the way, in that case I can attempt to rework your script to be more compatible; is it OK if I borrow some ideas from the FindSDL2 script?

d235j:
@rpavlik:
I've taken your code, fixed the regex (I think), and changed it so that while it still uses an imported target for SDL2::SDL2 and SDL2::SDL2main, it no longer uses an imported target for Cocoa and instead uses the IMPORTED_LINK_INTERFACE_LIBRARIES property.
This works because at this point, only system libraries use .tbd text-based stubs, so SDL2 will link fine directly. I think that until CMake provides a better way of doing this (and it looks like one is in development, and will be IMPORTED_LINK_ITEM), this is probably the best we can do.
I will open a pull request shortly.

PR #234, opened 2015-10-09
imaging merge (in-process memory image message and jni bridge code)

JeroMiya:
Optional in-process memory image message implementation (turn on with CMAKE option), and temporarily adding jni bridge hack to pluginKit.

### DIFF
--- a/src/osvr/Common/CMakeLists.txt
+++ b/src/osvr/Common/CMakeLists.txt
@@ -23,8 +23,11 @@ if(BUILD_WITH_TRACING)
     endif()
 endif()
 
+option(OSVR_COMMON_IN_PROCESS_IMAGING OFF)
+
 configure_file(TracingConfig.h.cmake_in "${CMAKE_CURRENT_BINARY_DIR}/TracingConfig.h")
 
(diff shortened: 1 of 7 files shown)


src/osvr/Common/CMakeLists.txt
```diff
@@ -23,8 +23,11 @@ if(BUILD_WITH_TRACING)
     endif()
 endif()
 
+option(OSVR_COMMON_IN_PROCESS_IMAGING OFF)
```
rpavlik: Think this is not quite the right syntax for `option()` - think there's usually a description in there.

src/osvr/Common/ImagingComponent.cpp
```diff
@@ -108,6 +108,47 @@ namespace common {
...
+                OSVR_ChannelCount sensor;
+                int buffer;
+            };
+            template <typename T>
+            void process(InProcessMemoryMessage &ipmmMsg, T &p) {
+                process(ipmmMsg.metadata, p);
+                p(ipmmMsg.sensor);
+                p(ipmmMsg.buffer);
+            }
+        }
```
rpavlik: `// namespace` closing comment

src/osvr/Common/CMakeLists.txt
rpavlik: also probably want a `mark_as_advanced`

PR #235, opened 2015-10-10
Implement getSerialDeviceList for Mac OS X

d235j:
Here is an implementation of getSerialDeviceList for Mac OS X. I have tested this outside of OSVR with a USB-serial adapter and it does function correctly. However, I have not been able to test with OSVR, but I will be doing so soon.

### DIFF
--- a/src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
+++ b/src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
@@ -61,24 +61,22 @@ namespace usbserial {
          * @return 0 on success, -1 on failure
          */
         int findSerialPorts(io_iterator_t *matchingServices) {
-
-            kern_return_t kernResult;
             // Query IOKit for services matching kIOSerialBSDServiceValue
             CFMutableDictionaryRef classesToMatch =
                 IOServiceMatching(kIOSerialBSDServiceValue);
             if (classesToMatch == NULL) {
-                return -1;
+                return false;
             }
             // Query only for serial ports
             CFDictionarySetValue(classesToMatch, CFSTR(kIOSerialBSDTypeKey),
                                  CFSTR(kIOSerialBSDAllTypes));
             // Run the query
-            kernResult = IOServiceGetMatchingServices(
+            kern_return_t kernResult = IOServiceGetMatchingServices(
                 kIOMasterPortDefault, classesToMatch, matchingServices);
             if (KERN_SUCCESS != kernResult) {
-                return -1;
+                return false;
             }
-            return 0;
+            return true;
         }
 
         /**
@@ -113,7 +111,7 @@ namespace usbserial {
         io_iterator_t serialPortIterator;
         io_object_t serialPortService;
         // find all serial ports
-        if (findSerialPorts(&serialPortIterator) == 0) {
+        if (findSerialPorts(&serialPortIterator) == true) {
             // iterate over serial ports, getting vid and pid
             while ((serialPortService = IOIteratorNext(serialPortIterator)) !=
                    0) {
--- a/src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
+++ b/src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
@@ -37,9 +37,9 @@
 #include <string>
 #include <fstream>
 
-#include <fcntl.h>        // for O_NONBLOCK
-#include <sys/ioctl.h>    // for ioctl
-#include <unistd.h>       // for open, close
+#include <fcntl.h>     // for O_NONBLOCK
+#include <sys/ioctl.h> // for ioctl
+#include <unistd.h>    // for open, close
 
 // System includes
 #include <CoreFoundation/CoreFoundation.h>
@@ -54,31 +54,53 @@ namespace usbserial {
     namespace {
 
         /**
-         * Helper function that returns an IOKit iterator for all serial ports on the system.
-         * @param matchingServices pointer which will be set to the value of the iterator
+         * Helper function that returns an IOKit iterator for all serial ports
+         * on the system.
+         * @param matchingServices pointer which will be set to the value of the
+         * iterator
          * @return 0 on success, -1 on failure
          */
         int findSerialPorts(io_iterator_t *matchingServices) {
 
             kern_return_t kernResult;
-            CFMutableDictionaryRef classesToMatch;
             // Query IOKit for services matching kIOSerialBSDServiceValue
-            classesToMatch = IOServiceMatching(kIOSerialBSDServiceValue);
+            CFMutableDictionaryRef classesToMatch =
+                IOServiceMatching(kIOSerialBSDServiceValue);
             if (classesToMatch == NULL) {
                 return -1;
             }
-            else {
-                // Query only for serial ports
-                CFDictionarySetValue(classesToMatch,
-                                     CFSTR(kIOSerialBSDTypeKey),
-                                     CFSTR(kIOSerialBSDAllTypes));
-            }
+            // Query only for serial ports
+            CFDictionarySetValue(classesToMatch, CFSTR(kIOSerialBSDTypeKey),
+                                 CFSTR(kIOSerialBSDAllTypes));
             // Run the query
-            kernResult = IOServiceGetMatchingServices(kIOMasterPortDefault, classesToMatch, matchingServices);
-            if(KERN_SUCCESS !=kernResult) {
+            kernResult = IOServiceGetMatchingServices(
+                kIOMasterPortDefault, classesToMatch, matchingServices);
+            if (KERN_SUCCESS != kernResult) {
                 return -1;
             }
             return 0;
+        }
+
+        /**
+         * Predicate that determines if the given vendor and product IDs match
+         * that of the provided USB serial device. If the @param vendorID or
+         * @param productID is @c boost::none, that parameter will match and ID.
+         *
+         * @param device the USB serial device to compare against
+         * @param vendorID optional vendor ID
+         * @param productID optional product ID
+         *
+         * @return true if the vendor and product IDs match, false otherwise
+         */
+        bool matches_ids(const USBSerialDevice &device,
+                         const boost::optional<uint16_t> &vendorID,
+                         const boost::optional<uint16_t> &productID) {
+            const bool vendor_matches =
+                (!vendorID || *vendorID == device.getVID());
+            const bool product_matches =
+                (!productID || *productID == device.getPID());
+
+            return (vendor_matches && product_matches);
         }
     }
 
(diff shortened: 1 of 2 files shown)


rpavlik:
Looks nice! I'd ideally like that character buffer to use a std::vector<char> instead of `malloc` and `free` (to be exception safe and RAII lifetime management), and the C-style casts to become C++-style-ones.

godbyk:
I'll do a code review on this shortly. I have to dig up a USB serial device again to test it.

d235j:
I've updated the PR to use C++-style casts and a vector (why did I forget about that...)

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
+        int findSerialPorts(io_iterator_t *matchingServices) {
+
+            kern_return_t kernResult;
+            CFMutableDictionaryRef classesToMatch;
+            // Query IOKit for services matching kIOSerialBSDServiceValue
+            classesToMatch = IOServiceMatching(kIOSerialBSDServiceValue);
+            if (classesToMatch == NULL) {
+                return -1;
+            }
+            else {
```
godbyk: The `else` keyword should be kept on the same line as the closing brace. But in this case, you don't need the `else` block since the `if` block just returns.

d235j: Thanks! Any other issues, so I get get them done all together?

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
-        boost::optional<USBSerialDevice>
-        make_USBSerialDevice(const std::string &device) {
-            //FIXME: STUB
-            return boost::none;
+        int findSerialPorts(io_iterator_t *matchingServices) {
+
+            kern_return_t kernResult;
+            CFMutableDictionaryRef classesToMatch;
+            // Query IOKit for services matching kIOSerialBSDServiceValue
+            classesToMatch = IOServiceMatching(kIOSerialBSDServiceValue);
```
godbyk: Combine the declaration on line 64 with the definition on line 66: `CFMutableDictionaryRef classesToMatch = IOServiceMatching(kIOSerialBSDServiceValue);`

In general, try to limit the scope of variables as much as possible.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
+            }
+            else {
+                // Query only for serial ports
+                CFDictionarySetValue(classesToMatch,
+                                     CFSTR(kIOSerialBSDTypeKey),
+                                     CFSTR(kIOSerialBSDAllTypes));
+            }
+            // Run the query
+            kernResult = IOServiceGetMatchingServices(kIOMasterPortDefault, classesToMatch, matchingServices);
+            if(KERN_SUCCESS !=kernResult) {
```
godbyk: Use spaces after the `if`, `for`, `while`, etc. keywords. Also binary operators should have a single space on either side.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
                         boost::optional<uint16_t> productID) {
 
-        //FIXME: STUB
         std::vector<USBSerialDevice> devices;
+
+        io_iterator_t serialPortIterator;
+        io_object_t serialPortService;
+        // find all serial ports
+        if(findSerialPorts(&serialPortIterator) != 0) {
+            goto bailout;
```
godbyk: Just `return devices;` here and drop the `goto` statement. Early `return`s are fine, `goto`s less so.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
+        }
+
+        // iterate over serial ports, getting vid and pid
+        while((serialPortService = IOIteratorNext(serialPortIterator)) != 0) {
```
godbyk: We could drop the `!= 0` here and get the same semantics, but I'm not sure what @rpavlik's take on that is. The way it's written now make it clear.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
+            const CFNumberRef vidObj = static_cast<CFNumberRef> (IORegistryEntrySearchCFProperty(serialPortService
```
godbyk: No space between `>` and `(`: `static_cast<Type>(value)`.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
+                                                                                 , kIOServicePlane
+                                                                                 , CFSTR("idProduct")
+                                                                                 , NULL
+                                                                                 , kIORegistryIterateRecursively | kIORegistryIterateParents));
+            const CFStringRef bsdPathObj = static_cast<CFStringRef> (IORegistryEntryCreateCFProperty(serialPortService,
+                                                                          CFSTR(kIOCalloutDeviceKey),
+                                                                          kCFAllocatorDefault,
+                                                                          0));
+
+            if(vidObj != NULL && pidObj != NULL && bsdPathObj != NULL) {
```
godbyk: Space after `if`. Drop the `!= NULL1 bits.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
+                CFNumberGetValue(vidObj, kCFNumberSInt16Type, &vid);
+                CFNumberGetValue(pidObj, kCFNumberSInt16Type, &pid);
+
+                // printf("ALL DATA: VID: %d, PID: %d, PATH: %s\n", vid, pid, bsdPathBuf);
+                if(vid == vendorID && pid == productID) {
+                    // convert the string object into a C-string
+                    CFIndex bufferSize = CFStringGetMaximumSizeForEncoding(CFStringGetLength(bsdPathObj), kCFStringEncodingMacRoman) + sizeof('\0');
+                    std::vector<char> bsdPathBuf(bufferSize);
+                    CFStringGetCString(bsdPathObj, &bsdPathBuf[0]
+                        , bufferSize, kCFStringEncodingMacRoman);
```
godbyk: Comma should be on the previous line.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
+                    // create the device
+                    USBSerialDevice usb_serial_device(vid,
+                                                      pid,
+                                                      &bsdPathBuf[0], &bsdPathBuf[0]);
+                    devices.push_back(usb_serial_device);
+                }
+            }
+        }
+
+    bailout:
```
godbyk: Drop the label after the `goto` above has been removed.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
+                return -1;
+            }
+            else {
+                // Query only for serial ports
+                CFDictionarySetValue(classesToMatch,
+                                     CFSTR(kIOSerialBSDTypeKey),
+                                     CFSTR(kIOSerialBSDAllTypes));
+            }
+            // Run the query
+            kernResult = IOServiceGetMatchingServices(kIOMasterPortDefault, classesToMatch, matchingServices);
```
godbyk: Write `kern_return_t kernResults = ...`

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
+         * @param matchingServices pointer which will be set to the value of the iterator
+         * @return 0 on success, -1 on failure
          */
-        boost::optional<USBSerialDevice>
-        make_USBSerialDevice(const std::string &device) {
-            //FIXME: STUB
-            return boost::none;
+        int findSerialPorts(io_iterator_t *matchingServices) {
+
+            kern_return_t kernResult;
```
godbyk: Remove this declaration.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,87 @@ namespace usbserial {
...
-         * @return an optional USBSerialDevice
+         * Helper function that returns an IOKit iterator for all serial ports on the system.
+         * @param matchingServices pointer which will be set to the value of the iterator
+         * @return 0 on success, -1 on failure
          */
-        boost::optional<USBSerialDevice>
-        make_USBSerialDevice(const std::string &device) {
-            //FIXME: STUB
-            return boost::none;
+        int findSerialPorts(io_iterator_t *matchingServices) {
```
godbyk: If the return value is only used for OK/error, then consider using a `bool` instead.

rpavlik:
Hint: for the little picky stuff, just run clang-format :) there are instructions in the repo.

d235j:
I ran the whole thing through clang-format; hopefully it's all right.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,89 @@ namespace usbserial {
...
+
+            if (vidObj && pidObj && bsdPathObj) {
+                // handle device
+                uint16_t vid, pid;
+                CFNumberGetValue(vidObj, kCFNumberSInt16Type, &vid);
+                CFNumberGetValue(pidObj, kCFNumberSInt16Type, &pid);
+
+                // printf("ALL DATA: VID: %d, PID: %d, PATH: %s\n", vid, pid,
+                // bsdPathBuf);
+                if (vid == vendorID && pid == productID) {
```
godbyk: We should allow for `boost::none` which should match any vendor or product ID.

d235j: Any any non-USB serial device as well?

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -51,51 +54,89 @@ namespace usbserial {
...
+                static_cast<CFNumberRef>(IORegistryEntrySearchCFProperty(
+                    serialPortService, kIOServicePlane, CFSTR("idProduct"),
+                    NULL,
+                    kIORegistryIterateRecursively | kIORegistryIterateParents));
+            const CFStringRef bsdPathObj =
+                static_cast<CFStringRef>(IORegistryEntryCreateCFProperty(
+                    serialPortService, CFSTR(kIOCalloutDeviceKey),
+                    kCFAllocatorDefault, 0));
+
+            if (vidObj && pidObj && bsdPathObj) {
```
godbyk: I would probably invert this conditional and loop earlier. This prevents deep nesting. `if (!vidObj || !pidObj || !bsdPatchObj) continue;`

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
godbyk: Just USB serial devices.

d235j: OK — that will require extra logic.

godbyk: See the original `matches_ids()` function above (line 80).

d235j:
I have addressed these issues; are there any remaining ones?

godbyk:
If you push the latest commits to this branch/PR, I'll take another look.

d235j:
I force-pushed them earlier. (Sorry, I know that hides older work.)

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -37,12 +37,15 @@
 #include <string>
 #include <fstream>
 
-#include <fcntl.h>        // for O_NONBLOCK
-#include <sys/ioctl.h>    // for ioctl
-#include <unistd.h>       // for open, close
+#include <fcntl.h>     // for O_NONBLOCK
+#include <sys/ioctl.h> // for ioctl
+#include <unistd.h>    // for open, close
```
godbyk: Do we still need these headers for the OS X implementation?

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -87,15 +85,62 @@ namespace usbserial {
...
+                CFNumberGetValue(vidObj, kCFNumberSInt16Type, &vid);
+                CFNumberGetValue(pidObj, kCFNumberSInt16Type, &pid);
+
+                // convert the string object into a C-string
+                CFIndex bufferSize = CFStringGetMaximumSizeForEncoding(
+                                         CFStringGetLength(bsdPathObj),
+                                         kCFStringEncodingMacRoman) +
+                                     sizeof('\0');
+                std::vector<char> bsdPathBuf(bufferSize);
+                CFStringGetCString(bsdPathObj, &bsdPathBuf[0], bufferSize,
```
rpavlik: `&bsdPathBuf[0]` can actually be more cleanly `bsdPathBuf.data()`

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -87,15 +85,62 @@ namespace usbserial {
...
+                                   kCFStringEncodingMacRoman);
+                // create the device
```
rpavlik: Does this guarantee that the buffer is null terminated?

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
d235j: I'm aware of that, just wasn't sure if we need to maintain C++03 compatibility.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
d235j: Yes:
https://developer.apple.com/library/mac/documentation/CoreFoundation/Reference/CFStringRef/#//apple_ref/c/func/CFStringGetCString
    The buffer must be large enough to contain the converted characters and a NUL terminator. For example, if the string is Toby, the buffer must be at least 5 bytes long.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
rpavlik: `.data()` is C++11? The things I forget.  No, not in this or any internal library - if it builds on VS2013 and some sampling of linux and mac compilers :D. (public APIs are C, with a C++03 header-only wrapper using Boost, that at least on the client end will probably be replaced by c++11 header-only wrappers)

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
rpavlik: That's annoying documentation - they heavily imply that the output gets null-terminated, but don't explicitly say so. Ah well, par for the course for string functions - for instance <https://randomascii.wordpress.com/2013/04/03/stop-using-strncpy-already/>

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
rpavlik: or returning your return value, and throwing in case of error.

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
d235j: We can look at the implementation in this case: https://github.com/opensource-apple/CF/blob/master/CFString.c#L2169

src/osvr/USBSerial/USBSerialDevInfo_MacOSX.h
```diff
@@ -87,15 +85,62 @@ namespace usbserial {
...
 
-        //FIXME: STUB
         std::vector<USBSerialDevice> devices;
+
+        io_iterator_t serialPortIterator;
+        io_object_t serialPortService;
+        // find all serial ports
+        if (findSerialPorts(&serialPortIterator) == true) {
+            // iterate over serial ports, getting vid and pid
+            while ((serialPortService = IOIteratorNext(serialPortIterator)) !=
```
godbyk: If `IOIternatorNext(nullptr)` is well-defined, then this could be written as `io_iterator_t serialPortIterator = findSerialPorts(); while ((serialPortService = IOIteratorNext(serialPortIterator)) { ... }` if you adjust the `findSerialPorts()` function to return the iterator (and change any `return false;` statements to `return nullptr;`).

rpavlik:
I'm going to call that nit-picking, and while it's certainly a good change if reasonable, no reason to not merge this pull request.

godbyk:
It's fine to merge, though the return type on the find function needs to be fixed in any case.

rpavlik:
then you can fix it, Kevin - we've got travis-ci building pull requests on Mac now :) let's not scare away a contributor :D

d235j:
I suppose you could merge the find function into the getSerialDeviceList function; I did consider doing that.

rpavlik:
no worries, thanks for the contribution!

PR #236, opened 2015-10-10
Adjust FindSDL2 to hopefully be able to find the framework on Mac (take two)

d235j:
This makes the aforementioned changes in https://github.com/OSVR/OSVR-Core/pull/232#issuecomment-147109322 and works for me.

### DIFF
--- a/cmake/FindSDL2.cmake
+++ b/cmake/FindSDL2.cmake
@@ -56,15 +56,6 @@ endif()
 
 include(FindPackageHandleStandardArgs)
 
-find_path(SDL2_INCLUDE_DIR
-	NAMES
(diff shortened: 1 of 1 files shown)


rpavlik:
Aside from some minor formatting tweaks (I hate spaces vs tabs and wish editors would  just handle it automatically better than they do), this looks great, thanks!

I assume you've tried it locally and it works with SDL? Doesn't look like Travis has SDL installed.

d235j:
Yep, this works with SDL. I could add SDL to the Travis configs if you want.

rpavlik:
cool. Feel free to, the more stuff builds on CI, the better it serves as a check to development.

PR #241, opened 2015-10-15
Add getProjectionClippingPlanes to C++ API

d235j:
This adds getProjectionClippingPlanes to the C++ API. There is a osvrClientGetViewerEyeSurfaceProjectionClippingPlanes in the C API, but there isn't such a function in the C++ API, which is particularly annoying as SteamVR wants clipping plane values, not a projection matrix.

### DIFF
--- a/inc/osvr/ClientKit/Display.h
+++ b/inc/osvr/ClientKit/Display.h
@@ -100,6 +100,13 @@ namespace clientkit {
         OSVR_ViewportDimension height;
     };
 
+    struct ProjectionClippingPlanes {
+        double left;
+        double right;
+        double top;
+        double bottom;
+    };
+
     /// @brief Wrapper for a viewer, eye, and surface bound to a display config.
     /// DOES NOT provide lifetime management for the associated display config!
     class Surface {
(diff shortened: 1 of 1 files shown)


d235j:
Rather than passing pointers, wouldn't it be cleaner to return an array?

rpavlik:
Yes, returning a nicely named struct would be nice. Passing or returning a raw array by value really only works when it's inside a struct or you'll get a shallow copy, not to mention the name-decay (hence why there are so many structs containing an array in OSVR, so they act more like objects).  Unknown order of elements here makes an array potentially confusing. I think the only consumer of this right now is the WebVR patch for Firefox (plus the Managed-OSVR wrapper, but I don't think anyone's using that function of it), so if you want to pull-request a breaking change in the C API for this, I think it would probably be reasonable.  cc @mars979 and @JeroMiya  for their opinions on those two projects respectively. (Or you could return a struct just from the C++ wrapper, leaving the C API as-is: makes binding easier, fewer custom datatypes)

As a workaround for gaps in the C++ API wrappers (of which there are many on the pluginkit side and some on the clientkit side), they're always designed to provide access to what you need to use the C API for filling those gaps, of course, but a pull request to fill the gap for everyone is even better :)

Will wait to merge until I hear back on the C API.

inc/osvr/ClientKit/Display.h
```diff
@@ -155,6 +155,21 @@ namespace clientkit {
             }
         }
 
+        /// @brief Gets the clipping planes.
+        ///
+        /// @sa osvrClientGetViewerEyeSurfaceProjectionClippingPlanes
+        void getProjectionClippingPlanes(double *left, double *right,
+                                         double *bottom, double *top) const {
```
rpavlik: In the C++ API, if I can't just return a return value, the pattern is to take parameters by reference (and take their address inside the wrapper) - more C++-like.

d235j:
Fixed to use references. (Sorry, much more familiar with C than C++!)

JeroMiya:
Let's keep the C API the same for now. However, I do strongly prefer returning a struct as opposed to several by-reference arguments in the C++ API (and the managed code wrappers too). It really reduces boilerplate code, eliminating the temporary variables you tend to create for each of the by-reference arguments. It's also easier to read code written using this style of API.

Sorry to make you modify your PR again, but could you create a ```ProjectionClippingPlanes``` struct that ```getProjectionClippingPlanes``` returns an instance of? Thanks!

d235j:
There we go, hopefully this is OK.

PR #242, opened 2015-10-16
Work on improving the video-based tracking

rpavlik:
There's substantial refactoring work here in the DirectShow code used on Windows, including replacing non-thread-safe bools-as-fake-semaphores with real sync primitives that should even perform better (windows set/reset events) and encapsulating the code internally a bit better.

Improved the performance of going from the DirectShow driver to OpenCV images - mainly, flipping in place is needlessly expensive. (yes, I used a profiler to verify that.)

Also, new app built but not installed: `ViewTrackingCamera` that just shows a live preview of what the tracking camera (not the algorithm, just the camera) sees. Used to avoid regressions when working on directshow, also useful for testing hardware. It shows a variable stride (default: 1 in 3 frames, pass a different stride as an argument if you like) of the feed from the camera in a little OpenCV window, saves one of the first frames out to a PNG file, and prints a framerate for the capture to the terminal - should be roughly 100 FPS.

The code now automatically turns on "50Hz powerline" mode for the HDK camera, which is really just a weird name for the correct high-gain mode we want. This should make the software more robust, in case somebody goes tinkering with camera settings or someone's comes with the defaults wrong.

Some Windows COM-related utilities have been split out into a separate directory for use in other projects - those will probably turn into a submodule eventually.


d235j:
It seems like MS wants developers to use Media Foundation (introduced in Vista) going forward — which seems to be why that qedit.h header is no longer present.

Additionally, for cross-platform support, we'll either have to write platform-specific code, or use libuvc. I haven't yet looked into camera support, as core and SteamVR plugins are higher priority to me, but I'd like to see native support for both OSVR HDK and Rift DK2 cameras eventually.

rpavlik:
Yes, as you might be able to see, this initially used OpenCV (which internally used directshow), but for some reason OpenCV can't read from the HDK camera, while this directshow code could. (The actual algorithm runs on OpenCV types, we just need to abstract the image source a little better.) I imagine on other platforms libuvc is probably the way to go since there are some up-close-and-personal details specific to tracking cameras (on the HDK camera, it has a specific VID/PID, and needs the "powerline frequency" set to "50Hz" to enter the right high-gain mode, and I know the data from the DK2 camera is all scrambled - there's code in here to descramble it.) . I probably need to get a Linux box set up here, since USB forwarding to a VM is hit-or-miss, and see how the camera actually works there.

d235j:
@rpavlik what VM software are you using? USB forwarding seems to be mostly OK with VMWare; not sure about VirtualBox but I've had issues with it in the past, unfortunately. That said, something is preventing the USB camera from forwarding in VMWare here, and I have not looked at it because that's not immediate priority for me.

rpavlik:
VirtualBox, and it's hit or miss whether I can even get the HID tracker forwarded. It apparently has some webcam forwarding and I got that to work once with a desktop camera app in Linux, but haven't tried running the tracker.

russell-taylor:
There will be issues running the DK2 camera on Windows.  The descrambling is one, and it will be a pain.  The real problem will be getting the camera into external-trigger mode.  This requires sending some data on its internal CAN bus, which I don't see how to do through HIDAPI.  There is code to do this in Linux with a different library, but I don't know how to do it on Windows.  If someone wants to dig into that, I can hunt down and post my notes.

d235j:
> This requires sending some data on its internal CAN bus

Is this something done using over USB (with vendor-specific UVC commands)? What library is being used to do this on Linux?
By the way, does the OSVR HDK camera use any sort of triggering?

rpavlik:
Just that sync cable between the beltbox and the camera. The 50Hz powerline frequency thing is the closest to "funny business" needed (and it should default to that, and even if it gets changed it still works it's just super dim so tracking range drops way off), aside from the fact that OpenCV can't open the camera successfully on its own. Skype can, however, (yes, I've done meetings while wearing the HMD in side-by-side mode with the IR camera instead of my webcam) so I'll just chalk it up to the crufty state of video capture APIs on Windows.

Also potentially contributing to the difference: the fact that we appear to have, in this code, one of the few, if only, permissively-licensed independent implementations of DirectShow video capture - everybody else seems to use a variant of a descriptively-named "videoInput" library that at least at its origin page has fairly sketchy license details (but good enough for OpenCV: https://github.com/Itseez/opencv/blob/ddf82d0b154873510802ef75c53e628cd7b2cb13/modules/videoio/src/cap_dshow.cpp#L49 )

russell-taylor:
d235j, This is done over USB with vendor-specific commands of some type.  Oliver Kreylos has posted the Linux code that does this on Github at https://github.com/Doc-Ok/OpticalTracking

The OSVR HDK camera exports a trigger that is sent to the HDK to control when the LEDs flash.

russell-taylor:
Ryan, It bugs me that I couldn't figure out what the video code I posted was doing to make it work; it looks like it is doing pretty much the same things as the OpenCV code.  I was hoping to find the difference and then post a fix to OpenCV and then keep using it rather than DirectVideo.

rpavlik:
well, I think in the long run we're actually better off using this code - it's more finely tuned now than the OpenCV stuff, which is just using that vendored-in library of dubious origin. With this, we can actually enumerate cameras and choose the one we want, set parameters, etc. Yes, we need a better interface between the image-grabbing and the tracking because all the ifdefs make my eyes bleed, but esp. now that I've merged this, this is actually pretty nice code. (windows set/reset events are handy things)

PR #248, opened 2015-10-20
Rendermanager config

JeroMiya:
Adding RenderManagerConfig parsing and server string to osvrClient as an internal API. It will be used directly by the RenderManager API.

### DIFF
--- a/inc/osvr/Client/RenderManagerConfig.h
+++ b/inc/osvr/Client/RenderManagerConfig.h
@@ -26,10 +26,13 @@
 #ifndef INCLUDED_RenderManagerConfig_h_GUID_C8DA5781_5B6C_454A_B4FF_1DB50CBE3479
 #define INCLUDED_RenderManagerConfig_h_GUID_C8DA5781_5B6C_454A_B4FF_1DB50CBE3479
 
-// Required for DLL linkage on Windows
+// Internal Includes
 #include <osvr/Client/Export.h>
 #include <osvr/Util/UniquePtr.h>
 #include <osvr/Util/ClientOpaqueTypesC.h>
+
+// Library/third-party includes
+// - none
 
 // Standard includes
 #include <string>
@@ -41,56 +44,56 @@ namespace osvr {
 namespace client {
 
     class RenderManagerConfig;
-    typedef unique_ptr<RenderManagerConfig> RenderManagerConfigPtr;
+    typedef std::shared_ptr<RenderManagerConfig> RenderManagerConfigPtr;
     class RenderManagerConfigFactory {
     public:
         OSVR_CLIENT_EXPORT static RenderManagerConfigPtr
-            create(OSVR_ClientContext ctx);
+            createShared(OSVR_ClientContext ctx);
     };
 
     class RenderManagerConfig {
     public:
-        OSVR_CLIENT_EXPORT RenderManagerConfig(const std::string& render_manager_config);
+        OSVR_CLIENT_EXPORT RenderManagerConfig(const std::string& renderManagerConfig);
         OSVR_CLIENT_EXPORT RenderManagerConfig() {};
 
-        void OSVR_CLIENT_EXPORT parse(const std::string& render_manager_config);
-        void OSVR_CLIENT_EXPORT print() const;
+        OSVR_CLIENT_EXPORT void parse(const std::string& renderManagerConfig);
+        OSVR_CLIENT_EXPORT void print() const;
 
         /// Read the property information.
-        bool OSVR_CLIENT_EXPORT getDirectMode() const;
-        unsigned OSVR_CLIENT_EXPORT getDisplayIndex() const;
-        bool OSVR_CLIENT_EXPORT getDirectHighPriority() const;
+        OSVR_CLIENT_EXPORT bool getDirectMode() const;
+        OSVR_CLIENT_EXPORT uint32_t getDisplayIndex() const;
+        OSVR_CLIENT_EXPORT bool getDirectHighPriority() const;
 
-        bool OSVR_CLIENT_EXPORT getEnableTimeWarp() const;
-        bool OSVR_CLIENT_EXPORT getAsynchronousTimeWarp() const;
-        float OSVR_CLIENT_EXPORT getMaxMSBeforeVsyncTimeWarp() const;
-        float OSVR_CLIENT_EXPORT getRenderOverfillFactor() const;
+        OSVR_CLIENT_EXPORT bool getEnableTimeWarp() const;
+        OSVR_CLIENT_EXPORT bool getAsynchronousTimeWarp() const;
+        OSVR_CLIENT_EXPORT float getMaxMSBeforeVsyncTimeWarp() const;
+        OSVR_CLIENT_EXPORT float getRenderOverfillFactor() const;
 
-        unsigned OSVR_CLIENT_EXPORT getNumBuffers() const;
-        bool OSVR_CLIENT_EXPORT getVerticalSync() const;
-        bool OSVR_CLIENT_EXPORT getVerticalSyncBlockRendering() const;
+        OSVR_CLIENT_EXPORT std::size_t getNumBuffers() const;
+        OSVR_CLIENT_EXPORT bool getVerticalSync() const;
+        OSVR_CLIENT_EXPORT bool getVerticalSyncBlockRendering() const;
 
-        std::string OSVR_CLIENT_EXPORT getWindowTitle() const;
-        bool OSVR_CLIENT_EXPORT getWindowFullScreen() const;
-        int OSVR_CLIENT_EXPORT getWindowXPosition() const;
-        int OSVR_CLIENT_EXPORT getWindowYPosition() const;
-        unsigned OSVR_CLIENT_EXPORT getDisplayRotation() const;
-        unsigned OSVR_CLIENT_EXPORT getBitsPerColor() const;
+        OSVR_CLIENT_EXPORT std::string getWindowTitle() const;
+        OSVR_CLIENT_EXPORT bool getWindowFullScreen() const;
+        OSVR_CLIENT_EXPORT int32_t getWindowXPosition() const;
+        OSVR_CLIENT_EXPORT int32_t getWindowYPosition() const;
+        OSVR_CLIENT_EXPORT uint32_t getDisplayRotation() const;
+        OSVR_CLIENT_EXPORT uint32_t getBitsPerColor() const;
 
     private:
         bool m_directMode;
-        unsigned m_displayIndex;
+        uint32_t m_displayIndex;
         bool m_directHighPriority;
-        unsigned m_numBuffers;
+        std::size_t m_numBuffers;
         bool m_verticalSync;
         bool m_verticalSyncBlockRendering;
 
         std::string m_windowTitle;
         bool m_windowFullScreen;
-        int m_windowXPosition;
-        int m_windowYPosition;
-        unsigned m_displayRotation;
-        unsigned m_bitsPerColor;
+        int32_t m_windowXPosition;
+        int32_t m_windowYPosition;
+        uint32_t m_displayRotation;
+        uint32_t m_bitsPerColor;
 
         bool m_enableTimeWarp;
         bool m_asynchronousTimeWarp;
--- a/inc/osvr/Client/RenderManagerConfig.h
+++ b/inc/osvr/Client/RenderManagerConfig.h
@@ -40,22 +40,6 @@
 namespace osvr {
 namespace client {
 
-    class RenderManagerConfigParseException : public std::exception {
-    public:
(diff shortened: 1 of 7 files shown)


russell-taylor:
Assuming that it compiles and runs, this looks okay to me.  The Core is not my domain, though.  Would like to see this pulled in so we can go forward with the cleaner RenderManager branch.

Would like to see example config files added.

inc/osvr/Client/RenderManagerConfig.h
```diff
@@ -0,0 +1,104 @@
...
+#include <string>
+#include <iostream>
+#include <exception>
+#include <vector>
+
+namespace osvr {
+namespace client {
+
+    class RenderManagerConfig;
+    typedef unique_ptr<RenderManagerConfig> RenderManagerConfigPtr;
```
rpavlik: who holds this `unique_ptr`? They aren't safe across DLL boundaries because the deletion function is generated in the usage scope. (Compare to `shared_ptr` where the deleter is captured and stored alongside the raw pointer at creation)

inc/osvr/Client/RenderManagerConfig.h
```diff
@@ -0,0 +1,104 @@
...
+        OSVR_CLIENT_EXPORT static RenderManagerConfigPtr
+            create(OSVR_ClientContext ctx);
+    };
+
+    class RenderManagerConfig {
+    public:
+        OSVR_CLIENT_EXPORT RenderManagerConfig(const std::string& render_manager_config);
+        OSVR_CLIENT_EXPORT RenderManagerConfig() {};
+
+        void OSVR_CLIENT_EXPORT parse(const std::string& render_manager_config);
```
rpavlik: the export things can be first on the line. Params are usually camel case.

inc/osvr/Client/RenderManagerConfig.h
```diff
@@ -0,0 +1,104 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef INCLUDED_RenderManagerConfig_h_GUID_C8DA5781_5B6C_454A_B4FF_1DB50CBE3479
+#define INCLUDED_RenderManagerConfig_h_GUID_C8DA5781_5B6C_454A_B4FF_1DB50CBE3479
+
+// Required for DLL linkage on Windows
```
rpavlik: Oh, I know this is nitpicky, but what happened to the stock "internal/library/standard" comments here? They help ensure clean compilation and avoid unexpected/undesirable inter-header interactions.

inc/osvr/Client/RenderManagerConfig.h
```diff
@@ -0,0 +1,104 @@
...
+    public:
+        OSVR_CLIENT_EXPORT RenderManagerConfig(const std::string& render_manager_config);
+        OSVR_CLIENT_EXPORT RenderManagerConfig() {};
+
+        void OSVR_CLIENT_EXPORT parse(const std::string& render_manager_config);
+        void OSVR_CLIENT_EXPORT print() const;
+
+        /// Read the property information.
+        bool OSVR_CLIENT_EXPORT getDirectMode() const;
+        unsigned OSVR_CLIENT_EXPORT getDisplayIndex() const;
```
rpavlik: Preferable to either use an explicit typedef (if it's comparable to something in the public API) to an explicitly-sized int, or an explicitly-sized int. The exception is `std::size_t` which is OK (suggested, even, when that's what it's describing - it's a nice meaningful name) except in public API (where the fact that it varies in size between platforms frequently, vs somewhat-less-frequently for the other un-sized int types, make it a pain to FFI)

src/osvr/Client/RenderManagerConfig.cpp
```diff
@@ -0,0 +1,190 @@
...
+    RenderManagerConfig::RenderManagerConfig(const std::string& render_manager_config)
+    {
+        parse(render_manager_config);
+    }
+
+    void RenderManagerConfig::parse(const std::string& render_manager_config)
+    {
+      Json::Reader reader;
+      Json::Value root;
+      reader.parse(render_manager_config, root, false);
```
rpavlik: Could fail here, by returning false - there's a JsonHelper.h or something in common that wraps this in a friendly way (throwing an exception if it blows up, instead of an easy-to-ignore return code.)

src/osvr/Client/RenderManagerConfig.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+      m_directMode = root["render_manager_parameters"]["direct_mode"].asBool();
+      m_displayIndex = root["render_manager_parameters"]["direct_display_index"].asUInt();
+      m_directHighPriority = root["render_manager_parameters"]["direct_high_priority"].asBool();
```
rpavlik: prefer something like this:

```c++
auto & params = root["render_manager_parameters"];
m_directMode = params["direct_mode"].asBool();
m_displayIndex = params["direct_display_index"].asUInt();
m_directHighPriority = params["direct_high_priority"].asBool();
```

Now, there's no error checking anywhere here, and the code all assumes that every value is present. If you want to access something with a default value, `params.get("direct_mode", false).asBool();` is the general idea.

(And the schema is less than ideal: names aren't camel case (standard for javascript and thus json), minimal use of nesting, but that's less of a code problem than a config file design problem)

src/osvr/Client/RenderManagerConfig.cpp
```diff
@@ -0,0 +1,190 @@
...
+      m_numBuffers = root["render_manager_parameters"]["num_buffers"].asUInt();
+      m_verticalSync = root["render_manager_parameters"]["vertical_sync"].asBool();
+      m_verticalSyncBlockRendering = root["render_manager_parameters"]
+        ["vertical_sync_block_rendering"].asBool();
+      m_windowTitle = root["render_manager_parameters"]["window_title"].asString();
+      m_windowFullScreen = root["render_manager_parameters"]["window_full_screen"].asBool();
+      m_windowXPosition = root["render_manager_parameters"]["window_x_position"].asInt();
+      m_windowYPosition = root["render_manager_parameters"]["window_y_position"].asInt();
+      m_displayRotation = root["render_manager_parameters"]["display_rotation"].asUInt();
+      m_bitsPerColor = root["render_manager_parameters"]["bits_per_color"].asBool();
```
rpavlik: Think the type is wrong here.

src/osvr/Client/RenderManagerConfig.cpp
```diff
@@ -0,0 +1,190 @@
...
+      m_directHighPriority = root["render_manager_parameters"]["direct_high_priority"].asBool();
+
+      m_numBuffers = root["render_manager_parameters"]["num_buffers"].asUInt();
+      m_verticalSync = root["render_manager_parameters"]["vertical_sync"].asBool();
+      m_verticalSyncBlockRendering = root["render_manager_parameters"]
+        ["vertical_sync_block_rendering"].asBool();
+      m_windowTitle = root["render_manager_parameters"]["window_title"].asString();
+      m_windowFullScreen = root["render_manager_parameters"]["window_full_screen"].asBool();
+      m_windowXPosition = root["render_manager_parameters"]["window_x_position"].asInt();
+      m_windowYPosition = root["render_manager_parameters"]["window_y_position"].asInt();
```
rpavlik: schema nitpicking - here I'd have made an object for window options, with a member for `fullscreen`, a member (holding an array of ints) called `position`, etc.

(I only nitpick because I still suffer under the inconsistent schema of the early display descriptors that got sucked into Unity before I had a chance to review it, and then changing it broke everybody's apps)

src/osvr/Client/RenderManagerConfig.cpp
```diff
@@ -0,0 +1,190 @@
...
+      m_displayRotation = root["render_manager_parameters"]["display_rotation"].asUInt();
+      m_bitsPerColor = root["render_manager_parameters"]["bits_per_color"].asBool();
+
+      m_enableTimeWarp = root["render_manager_parameters"]["time_warp"]["enable"].asBool();
+      m_asynchronousTimeWarp = root["render_manager_parameters"]["time_warp"]["asynchronous_time_warp"].asBool();
+      m_maxMSBeforeVsyncTimeWarp = root["render_manager_parameters"]["time_warp"]["max_ms_before_vsync"].asFloat();
```
rpavlik: Here I can actually show a bit better how my don't-repeat-yourself-accessor-style works, since there's actually something nested another level. The new scope is to limit the lifetime of the reference to just the code that needs it, keeps it clean and keeps the indentation matching the nesting level of the schema. (You can see a little of the cleanup I did on code that started like this in the display config stuff)

```c++
{
    auto & timeWarp = params["timeWarp"];
    m_enableTimeWarp = timeWarp["enable"].asBool();
    m_asynchronousTimeWarp = timeWarp["asynchronous"].asBool();
    m_maxMSBeforeVsyncTimeWarp = timeWarp["maxMillisecondsBeforeVsync"].asFloat();
}
```

src/osvr/Client/RenderManagerConfig.cpp
```diff
@@ -0,0 +1,190 @@
...
+      std::cout << "Window full screen: " << m_windowFullScreen << std::endl;
+      std::cout << "Window X Position: " << m_windowXPosition << std::endl;
+      std::cout << "Window Y Position: " << m_windowYPosition << std::endl;
+      std::cout << "Display rotation: " << m_displayRotation << std::endl;
+      std::cout << "Bits per color: " << m_bitsPerColor << std::endl;
+      std::cout << "Enable time warp: " << m_enableTimeWarp << std::endl;
+      std::cout << "Asynchronous time warp: " << m_asynchronousTimeWarp << std::endl;
+      std::cout << "Max ms before vsync time warp: " << m_maxMSBeforeVsyncTimeWarp << std::endl;
+      std::cout << "Render overfill factor: " << m_renderOverfillFactor << std::endl;
+    }
```
rpavlik: May just be easier to have the parser populate a struct than have a basically-read-only class with lots of accessors.

src/osvr/Client/RenderManagerConfig.cpp
```diff
@@ -0,0 +1,190 @@
...
+// Library/third-party includes
+#include <json/value.h>
+#include <json/reader.h>
+
+namespace osvr {
+namespace client {
+
+    RenderManagerConfigPtr RenderManagerConfigFactory::create(OSVR_ClientContext ctx) {
+        try {
+            auto const configString = ctx->getStringParameter("/display");
```
rpavlik: copy-paste error - wrong path. Ideally, stick a hardcoded path literal in a header somewhere so it's in only one place.

src/osvr/Client/RenderManagerConfig.cpp
```diff
@@ -0,0 +1,190 @@
...
+            RenderManagerConfigPtr cfg(new RenderManagerConfig(configString));
+            return cfg;
+        }
+        catch (std::exception const &e) {
+            OSVR_DEV_VERBOSE(
```
rpavlik: Probably shouldn't be verbose here, since most of the time there won't be a render manager config (none is being added by default, etc.) so the `getStringParameter` will probably throw, IIRC, otherwise if the config is slightly wrong, pretty sure jsoncpp will blow out on some of those accesses with an exception as well - those exceptions might be worth being verbose about - could split this into two try/catch sections.

src/osvr/Client/RenderManagerConfig.cpp
```diff
@@ -0,0 +1,190 @@
...
+                "Couldn't create a render manager config internally! Exception: "
+                << e.what());
+            return RenderManagerConfigPtr{};
+        }
+        catch (...) {
+            OSVR_DEV_VERBOSE("Couldn't create a display config internally! "
```
rpavlik: copy-paste typo

inc/osvr/Client/RenderManagerConfig.h
JeroMiya: I was following the pattern of <osvr/Client/DisplayConfig.h>. This type is the return value of the RenderManagerConfigFactory::create function. Should it be shared_ptr instead?

src/osvr/Client/RenderManagerConfig.cpp
JeroMiya: Where would be an appropriate place for this? Ideally it would be used by both src/osvr/Server/ConfigureServer.cpp and /src/osvr/Client/RenderManagerConfig.cpp. I didn't see one for the /display string (there's an internal one in Server), which is what this code was based off of.

rpavlik: A header in osvr/Common maybe - not critical, just convenient.

PR #251, opened 2015-10-28
Video developer info

russell-taylor:
This allows a config file to specify many more parameters to the algorithm.  The most significant for developers will be the ability to specify the patterns and positions of beacons for one or more sensors.  Modified the default config file to do this for the OSVR HDK.  Made a second config file that has all of the beacons (front and back) on one rigid sensor.  Wrote the developer documentation for how to add new sensors.


rpavlik:
The device descriptor (specifically, the semantic section) should be dynamically created based on the config passed, if nothing else to allow the user to name their tracked targets and allow the plugin to normalize the coordinate system.

Also, these docs, as well as the other ones recently added, should be linked in from the Doxygen documentation. See the doc directory for examples of how to do this. May have to add the directory to the doxyfile (the common one) for some markdown and image search path.

plugins/videobasedtracker/BeaconBasedPoseEstimator.cpp
```diff
@@ -202,11 +206,10 @@ namespace vbtracker {
         // do okay without using it, so leaving it out.
         // @todo Make number of iterations into a parameter.
```
rpavlik: this can be removed now, right?

plugins/videobasedtracker/BeaconBasedPoseEstimator.cpp
```diff
@@ -202,11 +206,10 @@ namespace vbtracker {
...
         bool usePreviousGuess = false;
-        int iterationsCount = 5;
         cv::Mat inlierIndices;
         cv::solvePnPRansac(
             objectPoints, imagePoints, m_cameraMatrix, m_distCoeffs, m_rvec,
-            m_tvec, usePreviousGuess, iterationsCount, 8.0f,
+            m_tvec, usePreviousGuess, m_solveIterations, 8.0f,
             static_cast<int>(objectPoints.size() - m_permittedOutliers),
```
rpavlik: can we get a meaning/name for that 8.0 magic number?

plugins/videobasedtracker/BeaconBasedPoseEstimator.h
```diff
@@ -66,12 +66,20 @@ namespace vbtracker {
         /// @param beacons 3D beacon locations
         /// @param requiredInliers How many "good" points must be available
         /// @param permittedOutliers How many additional "bad" points we can
+        /// @param solveIterations How many iterations to run the optimization algorithm for
+        /// @param maxReprojectionAxisError Maximum permitted reprojection error for a
+        ///        beacon that was identified in the image.  This is handled on a per-
+        ///        axis basis (it checks the maximum of X and Y reprojection error).
+        ///        It compares the image distance between the located beacon and the
+        ///        3D point reprojected into the image using the camera matrix.
         /// have
```
rpavlik: looks like this line belongs above

plugins/videobasedtracker/BeaconBasedPoseEstimator.h
```diff
@@ -66,12 +66,20 @@ namespace vbtracker {
         /// @param beacons 3D beacon locations
         /// @param requiredInliers How many "good" points must be available
         /// @param permittedOutliers How many additional "bad" points we can
+        /// @param solveIterations How many iterations to run the optimization algorithm for
+        /// @param maxReprojectionAxisError Maximum permitted reprojection error for a
+        ///        beacon that was identified in the image.  This is handled on a per-
+        ///        axis basis (it checks the maximum of X and Y reprojection error).
+        ///        It compares the image distance between the located beacon and the
+        ///        3D point reprojected into the image using the camera matrix.
```
rpavlik: to clarify, then, the units are camera pixels?

plugins/videobasedtracker/HDKLedIdentifier.cpp
```diff
@@ -108,26 +105,18 @@ namespace vbtracker {
...
+          if (d_patterns[i].find(bits) != std::string::npos) {
+            return static_cast<int>(i);
+          }
         }
 
         // No pattern recognized and we should have recognized one, so return
-        // a low negative.  We've used -2 so return -3.
+        // a low negative.  We've used -2 for too-small brightness
+        // difference so return -3 for this.
         return -3;
```
rpavlik: would be good to define these constants somewhere external since they show up in debug view

plugins/videobasedtracker/IdentifierHelpers.h
```diff
@@ -62,17 +62,18 @@ namespace vbtracker {
...
 
-        // Transform the brightnesses into a container of bools using this
-        // little lambda
+        // Transform the brightnesses into a string with '.' for dim
+        // and '*' for bright.
         std::transform(
             begin(brightnesses), end(brightnesses), begin(ret),
-            [threshold](Brightness val) { return val >= threshold; });
+            [threshold](Brightness val) {
+          if (val >= threshold) { return '*'; } else { return '.'; } });
```
rpavlik: can clang format this. Also, might be actually clearer with the ternary comparison operator: `return val >= threshold? '*' : '.';`

plugins/videobasedtracker/Types.h
```diff
@@ -78,6 +75,18 @@ namespace vbtracker {
...
     /// @}
+
+    /// Container for the information needed to define a sensor,
+    /// both the patterns of its beacons and their 3D spatial coordinates.
+    typedef struct {
+      PatternStringList patterns;
+      Point3Vector positions;
+      size_t requiredInliers;
+      size_t permittedOutliers;
+    } SensorDescription;
```
rpavlik: should just be struct SensorDescription, the typedefs are only needed in c-compilable code

plugins/videobasedtracker/com_osvr_VideoBasedHMDTracker.cpp
```diff
@@ -145,7 +148,7 @@ class VideoBasedHMDTracker : boost::noncopyable {
         m_currentImage = 0;
 
         if (m_images.size() == 0) {
-            std::cerr << "Could not read any images from " << VBHMD_FAKE_IMAGES
+            std::cerr << "com_osvr_VideoBasedHMDTracker: com_osvr_VideoBasedHMDTracker: Could not read any images from " << VBHMD_FAKE_IMAGES
```
rpavlik: assuming this double prefix is a typo?

plugins/videobasedtracker/com_osvr_VideoBasedHMDTracker.cpp
```diff
@@ -539,11 +574,63 @@ class ConfiguredDeviceConstructor {
     // Using `get` here instead of `[]` lets us provide a default value.
     int cameraID = root.get("cameraID", 0).asInt();
     bool showDebug = root.get("showDebug", false).asBool();
+    size_t solveIterations = root.get("solveIterations", 0).asInt();
+    double maxReprojectionAxisError = root.get("maxReprojectionAxisError", 0).asDouble();
+
+    // Default is an empty list, which can be filled in from the config
+    // file if we find entries there.
+    osvr::vbtracker::SensorDescriptionList mySensors;
+    if (root.isMember("sensors")) {
```
rpavlik: I'd suggest an early out on the opposite condition to reduce nested level.

plugins/videobasedtracker/com_osvr_VideoBasedHMDTracker.cpp
```diff
@@ -539,11 +574,63 @@ class ConfiguredDeviceConstructor {
...
+      Json::Value sensors = root["sensors"];
+      for (Json::ArrayIndex i = 0; i < sensors.size(); ++i) {
```
rpavlik: can actually use a range-based for-loop here: `for (auto const & sensor : sensors)` and simplify the subsequent code

plugins/videobasedtracker/com_osvr_VideoBasedHMDTracker.cpp
```diff
@@ -539,11 +574,63 @@ class ConfiguredDeviceConstructor {
...
+
+        Json::Value patterns = sensors[i]["patterns"];
+        Json::Value positions = sensors[i]["positions"];
+        if (patterns.size() != positions.size()) {
+          std::cerr << "com_osvr_VideoBasedHMDTracker: Mismatched pattern count ("
+            << patterns.size() << ") and position count ("
+            << positions.size() << ") in sensor description for sensor"
+            << i << std::endl;
+          return OSVR_RETURN_FAILURE;
+        }
```
rpavlik: Can the schema be reorganized to be (effectively) "an array of structs" (position and pattern pairs, though doesn't have to be an object-style thing, could just be a tuple-style array of two elements) instead of "struct of arrays"? would eliminate this class of errors and simplify this code.

plugins/videobasedtracker/BeaconBasedPoseEstimator.cpp
russell-taylor: Done

plugins/videobasedtracker/BeaconBasedPoseEstimator.cpp
russell-taylor: Done

plugins/videobasedtracker/BeaconBasedPoseEstimator.h
russell-taylor: fixed.

plugins/videobasedtracker/BeaconBasedPoseEstimator.h
russell-taylor: Clarified.  Yes, they are.

plugins/videobasedtracker/HDKLedIdentifier.cpp
russell-taylor: They are described in the Running.md document.  Added a description of what -2 means, since it was not described.

plugins/videobasedtracker/IdentifierHelpers.h
russell-taylor: I can never remember the order for the ternary comparison operator, so I tend not to use it.  Will go look up how to run clang-format.

plugins/videobasedtracker/Types.h
russell-taylor: That also compiled fine, so I changed it.

plugins/videobasedtracker/com_osvr_VideoBasedHMDTracker.cpp
russell-taylor: Yep, one too many ^V...

plugins/videobasedtracker/com_osvr_VideoBasedHMDTracker.cpp
russell-taylor: Except that I don't want out.  I still want to create using an empty sensor vector if there is not one in the file.

plugins/videobasedtracker/com_osvr_VideoBasedHMDTracker.cpp
russell-taylor: This nuked my index info in the error print, but let me take out some [i]s.  Also, I think auto is an abomination that makes it really hard to debug things, so I used the actual type there.  After looking at the new and old, decided to put back in the index-based so the error messages would tell which sensor had the problem.

plugins/videobasedtracker/com_osvr_VideoBasedHMDTracker.cpp
russell-taylor: Thought about doing it that way.  When used internally, it needs to be a pair of arrays.  Could swizzle it internally, but then the user also has to swizzle it on the way in and they probably already have it as two arrays so decided to leave it this way.  Put in the check to make sure they don't mess up the counts.  The complexity here simplifies elsewhere in the code and in the user's activity so I think this is the local minimum.

russell-taylor:
Okay, I think I've either rebutted or fixed everything except for moving the documentation to the right place, which I couldn't figure out how to do.  There are some more commits to look over.  Repeat a comment if you'd like to override my rebuttal and I'll do what's requested.

russell-taylor:
What's the status of this pull request?  Lots of good stuff in here going unused...

rpavlik:
Well, most of it no longer applies or has been manually merged or cherry
picked, I think basically all except the stuff where the whole beacon data
is in the config file. The plan is to eventually move away from that
particular plugin to the unified one (which shares some innards but is
structured differently primarily to handle IMU in the same filter as the
video data, and multiple targets per tracked body to get better performance
out of the rear target) if/once I get it to stop being so twitchy. (I do,
and should probably more frequently, by having a local checkout, refer to
it when working on unified)

The actual developer info itself should probably end up in OSVR docs.

On Thu, Apr 21, 2016, 7:33 AM Russell Taylor <notifications@github.com>
wrote:

> What's the status of this pull request? Lots of good stuff in here going
> unused...
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/OSVR/OSVR-Core/pull/251#issuecomment-212898045>
>
-- 
Ryan A. Pavlik, Ph.D.
Senior Software Engineer
Sensics, Inc.
www.sensics.com

PR #257, opened 2015-11-05
Use rotate_180 rather than right_roll in dSight portrait mode.

DuFF14:
right_roll property is ignored. Use rotate_180 instead.

### DIFF
--- a/apps/displays/Sensics_dSight_portrait_1input_sbs.json
+++ b/apps/displays/Sensics_dSight_portrait_1input_sbs.json
@@ -43,7 +43,7 @@
       {
         "center_proj_x": 0.5,
         "center_proj_y": 0.5,
-        "rotate_180": 180
+        "rotate_180": 1
       }
     ]
   }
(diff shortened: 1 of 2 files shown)


apps/displays/Sensics_dSight_portrait_2inputs.json
```diff
@@ -43,7 +43,7 @@
       {
         "center_proj_x": 0.5,
         "center_proj_y": 0.5,
-        "rotate_180": 0
+        "rotate_180": 180
```
rpavlik: should be 1 or true

apps/displays/Sensics_dSight_portrait_1input_sbs.json
```diff
@@ -43,7 +43,7 @@
       {
         "center_proj_x": 0.5,
         "center_proj_y": 0.5,
-        "rotate_180": 0
+        "rotate_180": 180
```
rpavlik: this is a bool masquerading as an int - don't recall exactly but pretty sure either 1 or true will do what you want. I know 1 will do what you want. I suspect 180 will not.

DuFF14:
Looks like 1 and 180 do the same thing. I can change it to 1 though if that's clearer. Still waiting to confirm the dsight configurations are correct.

rpavlik:
please do - then it matches the schema

PR #259, opened 2015-11-06
Tracker derivatives

rpavlik:
Adds capability to send and receive velocity and acceleration reports for trackers.

Could still use an example (printing velocity received from `/me/head` for instance) - not sure if that should just be added to an existing example or not.

Eventually should also get some helpers to convert from other angular velocity formats to the incremental rotation and dt - I have the formulas here.


JeroMiya:
This looks good. I was going to suggest using a different name instead of ReportTypesX.h, but I couldn't think of a better one.

russell-taylor:
An example would help us users know how to get this velocity information along with pose.  But I think go ahead and merge without it and mark it as a ToDo.

The only cooncerns I have are about the rotation, which I added as line comments.

rpavlik:
OK, did the rotations, so merging now. Examples can come later.

PR #260, opened 2015-11-06
Armada651 client nocv

rpavlik:
This removes the hard dependency on OpenCV - it's the continuation of #210. I just merged it with master locally, then did a few changes:

- some build stuff
- pulling the aligned alloc implementation into a cpp so we could provide a C API
- handling error conditions in the aligned alloc
- providing a C++ throwing wrapper for the aligned alloc
- providing a C++ allocation wrapper that returns a unique_ptr for great justice
- Additional updates to imaging component to always use our new functions (and to use them in the new code added to master since the branch point)


src/osvr/Common/ImagingComponent.cpp
```diff
@@ -231,12 +230,15 @@ namespace common {
...
+        /// of the unique_ptr that holds it now, rather than freeing?
+        auto imageBufferCopy = util::makeAlignedImageBuffer(imageBufferSize);
+        memcpy(imageBufferCopy.get(), imageData, imageBufferSize);
 
         Buffer<> buf;
-        messages::ImagePlacedInProcessMemory::MessageSerialization serialization(
-            messages::InProcessMemoryMessage{ metadata, sensor, reinterpret_cast<int>(imageBufferCopy) });
+        messages::ImagePlacedInProcessMemory::MessageSerialization
+            serialization(messages::InProcessMemoryMessage{
+                metadata, sensor, reinterpret_cast<int>(imageBufferCopy)});
```
rpavlik: @JeroMiya Would appreciate your help here, I'm sure I'm doing this wrong. The ownership semantics for that buffer aren't terribly clear in this in-process memory implementation, and I'm willing to bet that my guess was wrong, that the usage of `imageBufferCopy` should be `imageBufferCopy.release()` to take control back from the `unique_ptr` that now controls the allocated memory.

Also, `reinterpret_cast<int>(void *)` is terribly unsafe. I know this is for single-process usage, so we don't have to worry about differing pointer sizes on the wire, but at least use an `intptr_t` because `sizeof(int) != sizeof(void*)`

rpavlik:
Oh, and most of the "added" lines here aren't actually mine, I just moved some of @Armada651 's code around and GitHub doesn't show moved/renamed very smoothly.

rpavlik:
OK, @JeroMiya I realized I could just re-configure my local MSYS2/MinGW64 build quickly to be the in-process-imaging kind, so I fixed a typo I added, did the memory behavior that I think is right, and changed the message data type (the `intptr_t` thing - which was actually needed in order to compile on that platform: `sizeof(int) < sizeof(void*)` ). So now I just need a double-check that the logic is sound - I think it matches your previous logic but as I don't understand that code path as well, I'd like another opinion.

JeroMiya:
Looks good. The logic seems sound to me for the in-process-memory message changes as far as I can tell. I will just need to update the android imaging plugin, which I can do separately.

rpavlik:
@Armada651 Thanks again for doing the bulk of this work - it's much appreciated. Sorry it took so long to merge.

Armada651:
@rpavlik No problem, I'm just glad to see this getting merged. I'll work together with @feilen to replace boost with C++11 in the ClientKit which is the only other large dependency right now.

rpavlik:
Great, thanks a bunch. I did comment on his earlier patches, and did some improvements in master already to improve things like retrieving strings (removed a pointer usage by just using a vector buffer), so you can start there. If you need any MPL-like replacements, see the bundled TypePack headers (now merged, woohoo), though those shouldn't be necessary for just the C++ ClientKit API.

Armada651:
Ah I see you made AlignedMemory a C module? But it uses C++11 features, so that's why it doesn't compile with GCC.

feilen:
Cool to see this getting merged! The Boost dependencies have been stripped
off the first layer of headers in my branch, but I admittedly haven't had
the chance to dig further.

I'll put a few hours into it tonight, it shouldn't end up being a huge job,
just a little tedious.

src/osvr/Util/AlignedMemoryC.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+    // Reserve the first byte for our buffer pointer for later freeing.
+    void *aligned = (void **)buffer + 1;
+
+    // After this call the 'aligned' pointer will be aligned to a boundary.
+    // If there is not enough space to align the pointer, it stays
+    // unaligned.
+    std::align(alignment, bytes, aligned, space);
+
+    if (space < bytes || !aligned) {
```
Armada651: The `aligned` pointer will never be a `nullptr` here, the proper way to check if the alignment succeeded is to check the return value of `std::align`. If it succeeded it will return `aligned`, if not it will return `nullptr`.

See: http://en.cppreference.com/w/cpp/memory/align

rpavlik: Thanks - I fixed this later to not actually throw away the return value.

rpavlik:
(Yes, that Linux instance is really out of date - it's Wheezy.)

PR #263, opened 2015-11-07
Fix build with Ninja

d235j:
This allows building with ninja instead of make.
These changes are necessary because add_custom_command doesn't seem to work (likely because the -stamp names do not correspond to real output files), and add_custom_target both works and makes more sense here (again, as the -stamp names aren't real, so they can be considered targets).

ninja is significantly faster than make especially when doing rebuilds. It's worth trying out. To get cmake to generate ninja output, use -GNinja.

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -133,31 +133,32 @@ if(BUILD_SERVER_APP)
 
     # Copy contents of dir for both build and install trees.
     macro(osvr_copy_dir _dirname _glob _comment)
-        add_custom_command(OUTPUT "${_dirname}-stamp"
-            COMMAND "${CMAKE_COMMAND}" -E make_directory "$<TARGET_FILE_DIR:osvr_server>${_dirname}"
+        STRING(REPLACE "/" "-" targetname ${_dirname})
+        add_custom_target("${targetname}-stamp"
+            COMMAND "${CMAKE_COMMAND}" -E make_directory "$<TARGET_FILE_DIR:osvr_server>/${_dirname}"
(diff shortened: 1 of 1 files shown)


rpavlik:
I actually have been using Ninja (for Android and on Windows with MSYS2/MinGW), but as I know well by now, Ninja is a useful tool to experience race conditions. Will take a look at this.

rpavlik:
Is this related to CMake policy CMP0058? I've started getting that on my markdown-conversion outputs since updating msys2.

d235j:
This more than likely is related, but is not the same problem. ninja tries to write files named e.g. /displays-stamp, so the leading slash cannot be there. If I remove that, other things break, and in the end, this was the cleanest solution.

rpavlik:
My concern here is that Visual Studio slows down with every additional project, and each target turns into a project. If there's a way to keep everything as just a few targets, as before, that's preferable - it might just be possible to move the slashes out of the function argument into the body by its uses.  I do remember now why I did this weird stamp thing when it's generating files (that should obviate the need for fake stamps) - I couldn't use a generator expression in some particular place (custom target depends?) but I needed a generator expression in the destination of those file copies to handle MSVC, which makes subdirectories for compile types, etc.

I've been using Ninja here on Windows (MSYS2/MinGW) - what platform did you hit this ninja bug on so I can see if I can come up with alternate solutions?

d235j:
Platform was Mac OS X but I'd also expect this to happen on Linux.

rpavlik:
I'd hope so, or I'll have a hard time digging into this. Looking at it now.

d235j:
Fixed in #271.

PR #264, opened 2015-11-07
Put data files in share on non-Windows OS

d235j:
This changes the CMakeLists to put the data files in share/ on non-Windows operating systems. This may need some adjustment for Android, and I'm not sure if the location those files are placed in the build tree is ideal. This depends on https://github.com/OSVR/OSVR-Core/pull/263.

Issue #238, opened 2015-10-10
make install puts non-binaries in bin


d235j:
As currently structured, the CMakeLists put configuration files in bin. This is nonstandard.
We should instead have a share/osvr or share/osvr-core directory, and the sample configuration files, examples, and such go in there. (The docs should go in share/doc/osvr or share/doc/osvrcore, and this seems to currently be the case).

The default server config file should probably also go in the share folder.
Obviously, on Windows things might be different.

rpavlik:
Yep - this is a remnant revealing that Windows was the first "shipping" target platform (even though we built on Linux all along for portability and robustness testing) - see also #185, #171, and #172 among others.

Issue #185, opened 2015-08-06
config file paths


rpavlik:
On *nix, we should be using this: http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

Issue #172, opened 2015-07-08
Search path support for config files, json descriptors, etc.


rpavlik:
Right now, the server just looks in the working directory - which in the current typical use case (a Windows binary snapshot) is the `bin` directory. Instead of loading JSON files directly, it's reasonable that we might consider having a module in charge of loading json files that would handle paths, including per-user config paths (XDG standards on Linux, AppData on Windows, elsewhere on other platform) as well as varying system-wide paths (Linux should get config files in `/etc`, Windows you've got the program directory itself, ProgramData for system-wide stuff)

h/t @feilen for bringing this up and prompting filing it as an issue in GitHub rather than my brain.

PR #265, opened 2015-11-07
Automatically find config data or load default

d235j:
This enables automatic searching for config files based on a list of hardcoded paths (this can be changed if needed), or loads a default empty config if no config filename is passed.

This, together with https://github.com/OSVR/OSVR-Core/pull/264, should solve https://github.com/OSVR/OSVR-Core/issues/238, https://github.com/OSVR/OSVR-Core/issues/185, and https://github.com/OSVR/OSVR-Core/issues/171, https://github.com/OSVR/OSVR-Core/issues/172, at least, but may need some changes/tweaks.

Before we can claim full autoconfig support we also need to implement plugin directory path detection.

### DIFF
--- a/src/osvr/Server/CMakeLists.txt
+++ b/src/osvr/Server/CMakeLists.txt
@@ -11,12 +11,15 @@ set(API
 
 set(SOURCE
     ConfigureServer.cpp
+    ConfigureServerFromFile.cpp
     JSONResolvePossibleRef.h
     JSONResolvePossibleRef.cpp
     Server.cpp
     ServerImpl.cpp
     ServerImpl.h
     "${CMAKE_CURRENT_BINARY_DIR}/display_json.h")
+
+configure_file(ConfigFilePaths.h.cmake_in "${CMAKE_CURRENT_BINARY_DIR}/ConfigFilePaths.h")
 
 # Fallback display descriptor
 set(DISPLAY_JSON ../../../apps/displays/OSVR_HDK_1_1.json)
--- a/src/osvr/Server/ConfigFilePaths.h.cmake_in
+++ b/src/osvr/Server/ConfigFilePaths.h.cmake_in
@@ -1,0 +1,55 @@
+/** @file
+    @brief Auto-generated configuration header - do not edit!
+
+    @date 2015
+
+    @author
+    Sensics, Inc.
+    <http://sensics.com/osvr>
+*/
+
+// Copyright 2015 Sensics, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef INCLUDED_ConfigFilePaths_h_GUID_241E9C9C_0E0E_46B0_9DED_8F8059306192
+#define INCLUDED_ConfigFilePaths_h_GUID_241E9C9C_0E0E_46B0_9DED_8F8059306192
+
+#include <osvr/Util/PlatformConfig.h>
+namespace osvr {
+  namespace server {
+
+    std::vector<std::string> getDefaultConfigFilePaths() {
+
+#ifdef OSVR_WINDOWS
+        std::string appdata = std::getenv("APPDATA");
+#else
+        std::string home = std::getenv("HOME");
+#endif
+
+        std::vector<std::string> names;
+        names.push_back(home + "/.osvr/osvr_server_config.json");
+        names.push_back("@CMAKE_INSTALL_FULL_SYSCONFDIR@/osvr_server_config.json");
+
+#if defined(OSVR_WINDOWS)
+        names.push_back(appdata + "\\OSVR\\osvr_server_config.json");
+        names.push_back("osvr_server_config.json");
+#elif defined(OSVR_MACOSX)
+        names.push_back(home + "/Library/Application Support/OSVR/osvr_server_config.json");
+#endif
+
+        return names;
+    }
+  }
+}
+#endif // INCLUDED_ConfigFilePaths_h_GUID_241E9C9C_0E0E_46B0_9DED_8F8059306192
--- a/src/osvr/Server/ConfigFilePaths.h.cmake_in
+++ b/src/osvr/Server/ConfigFilePaths.h.cmake_in
@@ -27,7 +27,7 @@
 
 #include <osvr/Util/PlatformConfig.h>
 namespace osvr {
-  namespace server {
+namespace server {
 
     std::vector<std::string> getDefaultConfigFilePaths() {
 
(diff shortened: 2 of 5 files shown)


rpavlik:
Following up from discussion on #271 (moving discussion here where it fits better):

The displays directory is used in osvr_server config files: for example, this one https://github.com/OSVR/OSVR-Core/blob/master/apps/sample-configs/osvr_server_config.dSight.json  (The default is to prepopulate the data structure with a compiled-in version of the HDK display descriptor, so if you don't specify one explicitly that's what it uses.)

The "displays" entry can either contain the full display descriptor json object right there, or it can be a filename or a (limited subset of valid forms of) JSON reference. The code that handles that is in the following files, designed/coded with the knowledge it would probably expand, move into common or util, and gain awareness of some sense of "json search path" because right now it's pretty fragile and primarily current-working-directory-dependent (which happens to work OK in the case of Windows where working dir defaults to executable dir if you double-click it).

- header: https://github.com/OSVR/OSVR-Core/blob/master/src/osvr/Server/JSONResolvePossibleRef.h
- implementation: https://github.com/OSVR/OSVR-Core/blob/master/src/osvr/Server/JSONResolvePossibleRef.cpp

used for displays here: <https://github.com/OSVR/OSVR-Core/blob/master/src/osvr/Server/ConfigureServer.cpp#L345>
(It's used in a few other places in the server configuration procedure, like [resolving device descriptors for external VRPN devices](https://github.com/OSVR/OSVR-Core/blob/master/src/osvr/Server/ConfigureServer.cpp#L307), a [potentially separate rendermanager config file](https://github.com/OSVR/OSVR-Core/blob/master/src/osvr/Server/ConfigureServer.cpp#L368), etc.)

Eventually, of course, many times these display descriptors will be compiled in to plugins and  auto-detected, but there are plenty of cases/reasons when display parameters couldn't be autodetected or don't need a full plugin, so this configuration option is likely to stick around even once display autodetect functionality is in there.

src/osvr/Server/CMakeLists.txt
```diff
@@ -18,6 +19,8 @@ set(SOURCE
     ServerImpl.h
     "${CMAKE_CURRENT_BINARY_DIR}/display_json.h")
 
+configure_file(ConfigFilePaths.h.cmake_in "${CMAKE_CURRENT_BINARY_DIR}/ConfigFilePaths.h")
```
rpavlik: I don't think this file needs to be generated - it doesn't look like it's using any CMake substitutions. Just consuming `#ifdefs` from a configured header is fine in a regular header or implementation file.

src/osvr/Server/ConfigFilePaths.h.cmake_in
```diff
@@ -0,0 +1,57 @@
...
+    std::vector<std::string> getDefaultConfigFilePaths() {
+
+#ifdef OSVR_WINDOWS
+        std::string appdata = std::getenv("APPDATA");
+#else
+        std::string home = std::getenv("HOME");
+#endif
+
+        std::vector<std::string> names;
+        names.push_back(home + "/.osvr/osvr_server_config.json");
```
rpavlik: On *nix, possibly excluding Mac OS X, we should actually be using the base directories specified here: https://github.com/OSVR/OSVR-Core/issues/185

src/osvr/Server/ConfigFilePaths.h.cmake_in
```diff
@@ -0,0 +1,57 @@
...
+#define INCLUDED_ConfigFilePaths_h_GUID_241E9C9C_0E0E_46B0_9DED_8F8059306192
+
+#include <osvr/Util/PlatformConfig.h>
+namespace osvr {
+namespace server {
+
+    std::vector<std::string> getDefaultConfigFilePaths() {
+
+#ifdef OSVR_WINDOWS
+        std::string appdata = std::getenv("APPDATA");
```
rpavlik: I wrote a "friendly" wrapper for getenv at least in part to quiet Visual Studio about ownership and unify behavior if an environment variable is not set: https://github.com/OSVR/OSVR-Core/blob/master/inc/osvr/Common/GetEnvironmentVariable.h

src/osvr/Server/ConfigFilePaths.h.cmake_in
```diff
@@ -0,0 +1,57 @@
...
+        names.push_back(home + "/.osvr/osvr_server_config.json");
+        names.push_back(
+            "@CMAKE_INSTALL_FULL_SYSCONFDIR@/osvr_server_config.json");
+
+#if defined(OSVR_WINDOWS)
+        names.push_back(appdata + "\\OSVR\\osvr_server_config.json");
+        names.push_back("osvr_server_config.json");
+#elif defined(OSVR_MACOSX)
+        names.push_back(
+            home + "/Library/Application Support/OSVR/osvr_server_config.json");
```
rpavlik: Are these ever localized? Is there API to get their localized values?

src/osvr/Server/ConfigFilePaths.h.cmake_in
```diff
@@ -0,0 +1,57 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef INCLUDED_ConfigFilePaths_h_GUID_241E9C9C_0E0E_46B0_9DED_8F8059306192
+#define INCLUDED_ConfigFilePaths_h_GUID_241E9C9C_0E0E_46B0_9DED_8F8059306192
+
+#include <osvr/Util/PlatformConfig.h>
```
rpavlik: FYI, in case you hadn't seen it already, there's a boilerplate generator that will spit stuff out that looks like this (including the little headings for categories of include files) available here: http://tools.getosvr.org/boilerplate/  Please leave those headings in place, if possible, to help keep consistent style and organization.

A gist with some slick Windows-only integrations is here (puts it in your send-to menu), making a shell script variant (which is what this originally was before a PHP page) should be pretty easy: https://gist.github.com/rpavlik/fbdf1347e52b64b833c5

src/osvr/Server/ConfigFilePaths.h.cmake_in
d235j: Apple doesn't do localization this way. For localization, a directory gets named .localized, and a plist defines what the localized names should be. So no, /Library/Application Support never gets localized.

src/osvr/Server/ConfigFilePaths.h.cmake_in
d235j: I tried to do it manually (complete with GUID generation), but probably didn't get it completely right.

d235j:
Feel free to make any changes as needed. Also note that I didn't have a chance to test on Windows.

russell-taylor:
We may want to go ahead and put a default RenderManager config in place (see the installer for directmode one for the HDK).  This won't hurt clients that don't use RenderManager but would make it work by default without a special config file when they are.

rpavlik:
@russell-taylor we can do that in a separate issue, this one is primarily about cross-platforming stuff.

@d235j 

> Before we can claim full autoconfig support we also need to implement plugin directory path detection.

Not sure what you mean here - we do already have a plugin directory search path thing. It doesn't handle a system-wide or user-wide thing, but it does correctly handle (I think! I really need to set up a bare-metal Linux machine here) finding + loading plugins from the install tree.

d235j:
@rpavlik you're right, I should have noticed this, since I played with it before.

d235j:
Any thoughts on where we want to go with this? It may need some restructuring, but I'm not sure what exactly I should be doing.

JeroMiya:
@d235j hope you don't mind, I am in the process of rebasing this PR and addressing @rpavlik's feedback, as well as including some search path generalization code from @godbyk. I may need to push a new branch/PR and close this one, but I wanted to give you a heads up. Thanks for this work!

d235j:
@JeroMiya I don't mind, unfortunately I haven't had much of a chance to look at this stuff lately.

JeroMiya:
Cool, thanks. I'm going to go ahead and close this PR out then. It's being replaced by this one:
https://github.com/OSVR/OSVR-Core/pull/534

PR #268, opened 2015-11-11
Changed prediction interval for HDK from 1/60th of a second to 32ms, …

russell-taylor:
…which provides a more compelling experience in DirectMode.


rpavlik:
I'll wait to do this until after I've merged the vrpn pull request and updated the submodule, since I believe you tested with that updated velocity data.

russell-taylor:
Okay.  Yuval asked about getting this into master at the scrum this morning.

rpavlik:
OK, will do shortly.

PR #270, opened 2015-11-11
Client world transform

rpavlik:
Includes a number of improvements, including banishing Boost.Fusion, cleanup of remote handlers and their interaction with interfaces, and a client API to reset a client-context-local room-to-world transform based on the current yaw of the head. (See caveats of this initial iteration here: https://github.com/OSVR/OSVR-Core/commit/c378ce6ecbdf6d8be10ed2eb80b5f3b67f58cf6e )


rpavlik:
One question I have, is whether we need an example for this. I did a pretty basic modification to the OpenGL example (attached, had to rename to `.txt` to get GitHub to let me upload it here) to test it myself, but I don't really necessarily want to encourage it as part of "standard practice" or something you have to do, or complicate that example by making it show more things. On the other hand, if I commit this modified version of the example, that's essentially an identical file except for ~5 lines (and only about 2 of those are important: checking for the key, then calling the function - rest is just messaging).  Hoping perhaps just the doxygen docs are sufficient.

[OpenGLSample-modified-to-add-r-key.txt](https://github.com/OSVR/OSVR-Core/files/33129/OpenGLSample-modified-to-add-r-key.txt)

russell-taylor:
I feel the tension between wanting very simple examples and wanting examples of every feature.  RenderManager is kind of exploding with examples that all differ by only a small amount, each showing one feature.  I think having reset-yaw in this example program is not far enough down the slippery slope to worry about, given that it is doing it using the already-required cross-platform SDL library.  I'd say go ahead and add it to the example.

Hmm... I see your point about not wanting to encourage the use of the function, and having the main example include it would tend to do that.  Maybe a comment to the effect that this is how you would duplicate the specific Oculus function in the more-general room-to-world system within OSVR?

godbyk:
Perhaps instead of a huge proliferation of example programs we should have some cookbook-style documentation that can highlight the code snippets for doing common tasks (like reset-yaw) or to help folks transition from other APIs to OSVR. The recipes can then provide discussion and caveats about each function or pattern.

rpavlik:
Pretty sure there already is an example index - at least, I thought there was a doxygen file that had a comment to update it when you add new examples, though I can't find it offhand...

PR #271, opened 2015-11-12
Move the slash from the osvr_copy_dir arguments to its body.

rpavlik:
Fixes permission issues on Linux where Ninja tried to make /sample-configs with obvious lack of success.

This is my attempt to resolve the same issue as #263 without adding lots of targets (and thus terrifying/petrifying Visual Studio - the header dependency tests already do a good enough job of that)

@d235j could you test and see if this solves the ninja build on Mac?  I was able to reproduce the issue building with ninja on debian jessie, and this fixed it there.

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -134,30 +134,30 @@ if(BUILD_SERVER_APP)
     # Copy contents of dir for both build and install trees.
     macro(osvr_copy_dir _dirname _glob _comment)
         add_custom_command(OUTPUT "${_dirname}-stamp"
-            COMMAND "${CMAKE_COMMAND}" -E make_directory "$<TARGET_FILE_DIR:osvr_server>${_dirname}"
+            COMMAND "${CMAKE_COMMAND}" -E make_directory "$<TARGET_FILE_DIR:osvr_server>/${_dirname}"
             COMMENT "Making ${_comment} directory"
             VERBATIM)
         set_source_files_properties("${_dirname}-stamp" PROPERTIES SYMBOLIC TRUE)
         # Grab all the files with a glob, to avoid missing one.
-        file(GLOB _files RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}${_dirname}" "${CMAKE_CURRENT_SOURCE_DIR}${_dirname}/${_glob}")
+        file(GLOB _files RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/${_dirname}" "${CMAKE_CURRENT_SOURCE_DIR}/${_dirname}/${_glob}")
 
         foreach(FN ${_files})
             list(APPEND FILE_OUTPUTS "${_dirname}/${FN}")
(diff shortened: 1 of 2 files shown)


d235j:
Yes, this seems to work correctly on Mac OS X.
Please check out https://github.com/OSVR/OSVR-Core/pull/265 as well because that changes this function a bit to allow different destination paths.

d235j:
Weirdly I ended up not needing the add_dependencies(${LIBNAME_FULL} osvrUtil) after I cleaned things up. Based on CMake docs, target_link_libraries() should make add_dependencies() unnecessary...

rpavlik:
Great, glad to hear that it worked. The difference with add_dependencies/reason we need it here is because we don't want to start building Common (which uses a generated header in Util) until _after_ Util has finished building (thus guaranteeing that the file was generated), rather than letting Ninja build them in parallel and only blocking on link. (Yes, I had to look up add_dependencies to make sure it would do what I want.) The race loss only happens to me once in a while (and the generated header is fairly new), and only had it happen in a MinGW build.

I'll merge this then and look at re-working #265 accordingly.

d235j:
Figures...
The target_link_libraries(osvrUtilCpp) earlier is not good enough, I'm guessing?

rpavlik:
Right, I'm pretty sure that only specifies the dependency order for linking, rather than sequencing the entire Util build before anything in Common starts building.  Really, we just need to wait to build Common until after the custom command generating `ReportTypesX.h` executes, but I don't see how to do that other than waiting for the whole of Util to build, at which point we are guaranteed that the file has been created.

(I could be wrong, and/or we might want to adjust how the generation takes place, configure time vs. build time, but I think this is right for now. I put this at build time for one because it turns out generating hundreds of files at cmake time, as we used to do for the header-dependency tests, seriously slows down how long the generate step of CMake takes. If it proves too confusing, slow, and/or racing the way it is now, I am not at all opposed to reversing that choice.)

d235j:
I should have referenced https://cmake.org/pipermail/cmake/2007-August/015649.html earlier.

rpavlik:
Yeah, will look into it some more. That is an older post, however, and it's not strictly bringing up the case I ran into (the massively-parallel ninja, and having source for one target depend on a generated file from `add_custom_command` that is added to another target).  The third way, of course, is to have a separate target just for generating that file, that both libraries can depend on, but that feels messy or crude, for some reason...

I assume you meant re-working #264, because #265 doesn't actually mess with the build system. I was able to cherry-pick your #264 commit, with minor modifications (Windows install dir shouldn't be a generator expression, only build-time copy dir), quite easily. Does it make sense for me to just push my modified version of #264 before considering #265? I've got some comments/ideas on that one, though I would like to merge it soon.

d235j:
Sorry, I meant #264. I'd like your thoughts on that one too, regarding locations and whether they're OK. The goal is to no longer put data files in the bin directory except on Windows where you'd be only distributing the bin directory. Regarding #265, I'm wondering how the displays directory is used, as we need to have a way of querying for display parameters.

PR #273, opened 2015-11-13
Put data files in PREFIX/share on non-Windows platforms

rpavlik:
This is primarily just #264 by @d235j - I just cherry-picked, let kdiff auto-resolve conflicts, and tweaked Win behavior.

A quick thumbs-up/thumbs-down would be good to make sure I didn't break things in ways that Travis-CI isn't testing.

Closes some subset of the issues listed in #265.


d235j:
Works, though is $BUILDDIR/data a good place to put these files in the build tree? Or should they stay in bin? (share doesn't exist in the build tree...)

rpavlik:
Well, it doesn't actually matter - we could have share exist in the build tree, that would be fine. The main goal was just to make it possible to run from both a build tree and an install tree without much pain, and Visual Studio (slash multi-configuration generators - I have no idea how a modern XCode organizes its output files) is typically the stumbling block for that goal.

rpavlik:
I decided to have them copied in a build tree to a parallel location as they'd be in an install tree, to hopefully simplify search logic.

PR #280, opened 2015-11-19
Remove boost dependencies from ClientKit and anything it includes

feilen:
Sorry this took so long!

Possible issues:
- BoostDeletable.h is now StdDeletable.h. would 'SharedDeletable.h' or any other name make more sense?
- CMake didn't like it when I provided 'osvr_cxx11_flags' as the target_compile_features, so I was forced to use cxx_range_for. Works, but not ideal, if anyone can figure out why that is.

I dug for includes or anything else that included Boost to a depth of six, so hopefully I caught everything. Let me know if there's anything else missing and I'll try to clean it up.


inc/osvr/Util/StdDeletable.h
```diff
@@ -22,25 +22,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef INCLUDED_StdDeletable_h_GUID_2C03AF21_E4B8_44B3_0F77_630666063711
-#define INCLUDED_StdDeletable_h_GUID_2C03AF21_E4B8_44B3_0F77_630666063711
+#ifndef INCLUDED_Deletable_h_GUID_66D6C45A_7AB7_4198_F8EE_202F0FCE682C
+#define INCLUDED_Deletable_h_GUID_66D6C45A_7AB7_4198_F8EE_202F0FCE682C
```
Armada651: No need to change this header guard right?

Armada651:
I think we still need a commit that removes boost as a CMake dependency when only the ClientKit is being compiled.

feilen:
Ah, quite possibly. It appears as though I can't compile the current OSVR
branch of dolphin with or without these changes right now, due to system
lib changes, so testing will be a bit harder.

Is there a circumstance where we'd just want to compile the clientkit
though? Almost certainly what you'd have is OSVR installed system wide, and
just use its headers and precompiled clientkit lib.

inc/osvr/Util/StdDeletable.h
feilen: I've no idea, is that auto-generated?

I think clang-format is what is used for keeping things tidy, but I've just been trying to carefully keep in sync with the surrounding code.

feilen:
Brought Dolphin/OSVR up to date with master and compiled it on a system with this version of OSVR installed and no boost headers, built without issues!

I'll look into the CMakeLists.txt, it's probably best to use that for completeness.

(for documentation purposes: the plan is to pull OSVR-Core as a submodule into Dolphin's externals folder, and build just the clientkit as part of the Dolphin build)

feilen:
Hmm, it will take more than just this to do that though, as this only removes boost from the ClientKit includes... not clientkit itself.

For ClientKit itself, the following files would need to have boost references removed:

```
osvr/Client/InternalInterfaceOwner.h
osvr/Client/ViewerEye.h
osvr/Common/ClientContext.h
osvr/Common/ClientInterface.h
osvr/Common/DegreesToRadians.h
osvr/Common/GetEnvironmentVariable.h
osvr/Common/InterfaceState.h
osvr/TypePack/And.h
osvr/TypePack/Contains.h
osvr/TypePack/Fold.h
osvr/TypePack/If.h
osvr/TypePack/List.h
osvr/TypePack/Not.h
osvr/TypePack/Quote.h
osvr/TypePack/T.h
osvr/TypePack/TypeKeyedTuple.h
osvr/Util/Angles.h
osvr/Util/KeyedOwnershipContainer.h
osvr/Util/SharedPtr.h
osvr/Util/TimeValueC.h
```
Liiiittle bit bigger job.

inc/osvr/Util/StdDeletable.h
rpavlik: If you rename the file, it's courteous to rename the header guard. In this case, leave it intact.  There's a nifty (well, kind of hacky but useful) boilerplate generator that spits out files like that - think it might be linked to in the CONTRIBUTING file? or maybe HACKING? it's on like tools.getosvr.org and there are some powershell scripts to add it to your right-click send to menu

PR #284, opened 2015-11-20
Refactor vbtracker capture

rpavlik:
Cleans up the plugin substantially, by splitting out responsibility for setting up the image capture (real or fake) and image conversion from the actual code. Enables use of OpenCV to try the tracking on non-Windows (camera firmware v7 likely required, as that's what fixed OpenCV capture on Windows). Cleanly pulls out the DK2 camera stuff so it's both a) not cluttering up the code and b) easier to make use of when desired. Will also enable easier automated testing (since now the fake images can be presented exactly as if they were a live camera stream).

@d235j hopefully this is also useful to you in libuvc usage, there are `@todo` or `TODO` comments where you'd change it to use something other than OpenCV for capturing on non-Windows.


rpavlik:
The Travis failure is a build system thing, it just needs the CMake position independent flags to link those static libs into a shared lib, will look up the name of that variable Monday.

PR #293, opened 2015-11-25
osvr_server support for cmdline options and blank config generation (resolves #228)

### LINKED ISSUE #228: Configuration-free server
Since the default config file for `osvr_server` is now empty, why should `osvr_server` require a configuration file be present at all?  I propose the following behavior:
1. If a configuration file is provided on the command line, use it (if it exists).
2. Failing (1), if an `osvr_server_config.json` file is available in the current working directory use it.
3. Failing (1) and (2), start the server using an empty `{ }` configuration.

Outurnate:
Patch to resolve #228.  Includes options --config= and --help.  Will create config with contents "{ }" in the event that the given config doesn't exist

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -148,7 +148,7 @@ if(BUILD_SERVER_APP)
     add_executable(osvr_server
         osvr_server.cpp
         ${OSVR_SERVER_RESOURCE})
-    target_link_libraries(osvr_server osvrServer jsoncpp_lib)
+    target_link_libraries(osvr_server osvrServer boost_program_options boost_filesystem jsoncpp_lib)
     set_target_properties(osvr_server PROPERTIES
         FOLDER "OSVR Stock Applications")
     install(TARGETS osvr_server
--- a/apps/osvr_server.cpp
+++ b/apps/osvr_server.cpp
@@ -53,35 +53,47 @@ void handleShutdown() {
 
 int main(int argc, char *argv[]) {
     const std::string helpOpt("help");
-    const std::string configOpt("config");
+    const std::string configOpt("option");
 
     std::string configName;
 
-    opt::options_description options("OSVR Server Options");
-    options.add_options()(
+    opt::options_description optionsAll("All Options");
+    opt::options_description optionsVisible("Command Line Options");
+    opt::positional_options_description optionsPositional;
+
+    optionsPositional.add(configOpt.c_str(), -1);
+    optionsAll.add_options()(
         configOpt.c_str(),
         opt::value<std::string>(&configName)
             ->default_value(osvr::server::getDefaultConfigFilename()),
-        "override config filename")(helpOpt.c_str(),
-                                    "display this help message");
+        "override config filename");
+    optionsVisible.add_options()(helpOpt.c_str(), "display this help message");
+    optionsAll.add(optionsVisible);
 
     opt::variables_map values;
     try {
-        opt::store(opt::parse_command_line(argc, argv, options), values);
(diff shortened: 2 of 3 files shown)


godbyk:
Instead of `osvr_server --config=<config file>`, I think `osvr_server [config file]` would be better (where the config file is optional and defaults to an empty JSON string as you've done).

Outurnate:
Done!  Removed the line from help as well.

Quick error check was added to config file creation

godbyk:
Running `osvr_server --help` only shows the `--help` command-line option. It doesn't show the config file option.

Also, instead of creating a file in case someone happens to misspell it, it would be better to construct the server config from the empty JSON string. (This probably entails refactoring `osvr::server::configureServerFromFile()` into some smaller helper functions.)

Outurnate:
I think the chance of misspelling is low - it'll be daemonized by most users.  Either way, I agree, and the function should be split up

godbyk:
:+1: Looks good to me. Thanks, @Outurnate!

araujobsd:
There are some conventions for command line parameters, basically the usage of "-" or "--" or "/" makes more sense then only "osvr_server [config file]"

Personally I have strong objection about the update, and the previous patch was much better.

godbyk:
@araujobsd There are conventions for command-line parameters and I believe my comments reflect those conventions. See the fourth paragraph under [Standards for Command Line Interfaces](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html#Command_002dLine-Interfaces) from the GNU Coding Standards which also generally reflect the POSIX conventions.

araujobsd:
Yeap, I know that too! Thanks to point it out, my answer was based on POSIX.1-2008 ([Utility Conventions](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html)).

And my main concern is, if in the future we want to extend the options via CLI such like "example": osvr_server --iface=eth1 --ipv4=0.0.0.0 --config=config_file

... looks for me [-c|--config] is more consistent than "osvr_server config_file"; you can argue that all those options shall be passed via config_file and never via CLI, however file as an operand via CLI for my eyes looks strange. Maybe I'm being pedantic.

godbyk:
@araujobsd The POSIX doc you linked to agrees with my suggestion. The config filename is an [operand](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_255), to use their terminology.

I do envision `osvr_server` taking command-line options—especially for more transient settings such as setting log verbosity levels or to temporarily override a configuration file setting. But I also anticipate it taking multiple configuration files that it would compose together internally. This would allow plugins to ship plugin-specific configuration files and allow the user to tell the server to load a few of them simultaneously.

The syntax for `osvr_server` might then be something like:

    osvr_server [-v|--verbose] [-D|--daemon] [-V|--version] [config files...]

and an example execution might be:

    osvr_server --verbose my_custom_hmd.json razer_hydra.json omni_treadmill.json

This usage is much more common than, say, `osvr_server -c my_custom_hmd.json -c razer_hydra.json --config=omni_treadmill.json`. Options are rarely repeated (though there are some common exceptions like `-v`, `-vv`, etc. to increase verbosity).

araujobsd:
Hey,

OK, I didn't envision multiple configuration files at once in my previous replies, that is why I was pedantic with it. Actually you don't need `-c config1 -c config2 -c config3` nobody does that, but instead you can have `[-c|--config] config1 config2 config3`.

But makes sense why you choose that way after your explanation.

Thanks.

PR #296, opened 2015-11-30
Convert RenderManagerConfig.h to header-only

JeroMiya:
Fixes an issue where corrupted string data causes an exception in render manager when used from OSVR-Unreal. If this were a public API, the real fix would be a C API with a header-only C++ wrapper, but this is still an internal API.


JeroMiya:
Not sure what the ```ReportTypesX.h``` header test failure is in the ```continuous-integration/travis-ci/pr``` failure. @rpavlik any thoughts?

rpavlik:
that build thing is a sign I need to change that generation to config-time rather than build time, because that's a lovely little race condition that I can't apparently dodge.

PR #298, opened 2015-11-30
directx_samplegrabber_callback: Add missing DirectShow include.

Armada651:
Without this include I am not able to compile this module, it results in a ton of undefined identifier errors.


rpavlik:
What version of the Windows SDK are you using, out of curiosity?

Armada651:
For the qedit header I'm using 7.1A.

rpavlik:
Ah, sure, I apparently went overboard cleaning up includes. Not sure why it worked on my system. `BITMAPINFOHEADER` isn't defined by magic :)

PR #301, opened 2015-12-07
Vfov and hdk distortion

russell-taylor:
This modifies the vertical field of view for many of the HMDs to take into account the impact of the arctangent on the shift from angles to pixels, and the tangent when switching back.  This provides an improved viewing experience, reducing distortion.


DuFF14:
Core won't be able to parse the distortion parameters "polynomial_coeffs_X".
It is looking for "k1_X", where X = red, green, or blue.
DisplayDescriptorSchema needs to be updated to support polynomial distortion parameters: https://github.com/OSVR/OSVR-Core/blob/2cde03b2f6b176a70554e7ec981a5b4bcd10da2d/src/osvr/Client/DisplayDescriptorSchema1.h#L97
https://github.com/OSVR/OSVR-Core/blob/f10d9cbfe15bb26696daa60068ec9d67d9dd7dde/src/osvr/Client/DisplayDescriptorSchema1.cpp#L107

DuFF14:
I see this only applies to the new HDK 1.3 descriptor, so should be fine to merge.

rpavlik:
So is this safe to merge without breaking core-only apps?

apps/displays/OSVR_HDK_1_3.json
```diff
@@ -0,0 +1,52 @@
...
+  "meta": {
+    "schemaVersion": 1
+  },
+  "hmd": {
+    "device": {
+      "vendor": "OSVR",
+      "model": "HDK",
+      "num_displays": 1,
+      "Version": "1.2",
+      "Note": "Suitable for HDK 1.0-1.2"
```
rpavlik: Is this actually for 1.0-1.2 or for 1.3?

apps/displays/OSVR_HDK_1_3.json
```diff
@@ -0,0 +1,52 @@
...
+        "display_mode": "horz_side_by_side",
+        "swap_eyes": 0
+      }
+    ],
+    "distortion": {
+        "distance_scale_x": 1,
+        "distance_scale_y": 1,
+        "polynomial_coeffs_red": [ 0, 1, -0.95, 1.88, 1.62, 1.11 ],
+        "polynomial_coeffs_green": [ 0, 1, -0.95, 1.88, 1.62, 1.11 ],
+        "polynomial_coeffs_blue": [ 0, 1, -0.95, 1.88, 1.62, 1.11 ]
```
rpavlik: yeah, these are totally new fields. Are these parsed by render manager as-is, or is the schema flexible, since having parallel fields named this way is not great.. would be much better to have:

```json
"polynomialCoefficients": {
  "red": [ 0, 1, -0.95, 1.88, 1.62, 1.11 ],
  "green":  [ 0, 1, -0.95, 1.88, 1.62, 1.11 ],
  "blue": [ 0, 1, -0.95, 1.88, 1.62, 1.11 ]
}
```

apps/displays/OSVR_HDK_1_3.json
russell-taylor: Never looked at the comments, so they are apt to be aged.  The distortion correction makes the file suitable only for 1.3.

apps/displays/OSVR_HDK_1_3.json
russell-taylor: They are parsed by RenderManager as is.  Was using the same construction as before, which had K's for red,green,blue.  Not wedded to any particular format.  Same could be said of the distance_scale, which could have been a vector but I thought having X and Y explicit made sense.  Could nest X and Y of course.

Just need to know what it changes to if it changes so can adjust the parser.  Would like this all to be parsed in Core, of course, and sent over in a structure.  Jeremy wanted to get things running sooner, so we're in a middle state.

Note that in the immediate future, we're going to need to extend this to pass a (large) mesh, so we may want to hold off on pulling into core until that happens.

apps/displays/OSVR_HDK_1_3.json
russell-taylor: Also, the version should be 1.3 I guess.  Sorry about that.  Only changed what needed to be changed and did not look carefully at the context.

russell-taylor: Fixed in upcoming commit.

apps/displays/OSVR_HDK_1_3.json
russell-taylor: Per our HipChat discussion, I'm leaving this to be fixed along with the migration away from per-axis names generally.

russell-taylor:
Okay, fixed the Json file and it looks like we're waiting for the future to fix the per-axis naming that is going on.  Should be ready to merge.

PR #308, opened 2015-12-14
Resolved conflicts in stdInt while compiling in Visual Studio 2010

lexert:
stdint.h is available starting with Visual Studio 2010, so we must include it
because others headers will include it and cause conflicts
(at least redefenition error of int_fast16_t and uint_fast16_t)

### DIFF
--- a/inc/osvr/Util/StdInt.h
+++ b/inc/osvr/Util/StdInt.h
@@ -31,7 +31,7 @@
 
 /* IWYU pragma: begin_exports */
 
-#if !defined(_MSC_VER) || (defined(_MSC_VER) && _MSC_VER > 1600)
+#if !defined(_MSC_VER) || (defined(_MSC_VER) && _MSC_VER >= 1600)
 #include <stdint.h>
 #else
 #include "MSStdIntC.h"


rpavlik:
I'm assuming you're building client applications or plugins with MSVC 10, since the core itself will definitely not compile with MSVC 10. If you've gotten the stdint.h in msvc10 to work, then more power to you: if you just want to let me know that you've tried this and it works for your use case I'll gladly merge it.

lexert:
Yes, it perfectly works when building the client applications using vc10 compiler. And yes i knew that for compiling the core sdk the commit does not  make much sense

lexert:
I am using only a small subset of the client api though and had not tried yet writing any plugins, so theoretically here might be any problems in some parts of client api with compilation on vs10

rpavlik:
Sounds good. I had just in the past but been able to actually get stdint in vc10 to actually work, which is why I wanted the double check, that's all. Thanks a bunch for the contribution, hope to hear more from you! If you're not already, please join the dev discussion mailing list

rpavlik:
(I've actually built osvr apps using vc9, so yes, that's certainly an intended use case for the c API)

PR #314, opened 2015-12-23
Changed distortion coefficients on the HDK 1.3 to better ones for the…

russell-taylor:
… corrected field of view.


rpavlik:
Made a small change because the note field is actually printed in server startup. Ready to merge, though may as well wait until we merge my video tracker branch to trigger only one build

PR #320, opened 2016-01-09
CMake: Bring up Boost 1.60 incompatibility to front

bwrsandman:
Underline that osvr/Common/IPCRingBuffer.cpp is [incompatible](https://github.com/OSVR/OSVR-Core/blob/eeaf6ef/src/osvr/Common/IPCRingBuffer.cpp#L56) with Boost 1.60
Prevent CMake generation and compiling up to IPCRingBuffer which will fail

### DIFF
--- a/cmake-local/BoostTargets.cmake
+++ b/cmake-local/BoostTargets.cmake
@@ -18,6 +18,9 @@ if(NEED_BOOST_LOCALE)
 endif()
 
 find_package(Boost 1.44 COMPONENTS ${required_boost_components} REQUIRED) # Lower version bound of 1.43 for range adapters, 1.44 for Filesystem v3
+if(Boost_VERSION GREATER 105900)
+    message(SEND_ERROR "Using an untested Boost version - inspect the Boost Interprocess release notes/changelog to see if any ABI breaks affect us.")
+endif(Boost_VERSION GREATER 105900)
 
 if(NEED_BOOST_THREAD)
     add_library(boost_thread INTERFACE)


rpavlik:
If you can check that Boost 1.60 hasn't changed the ABI for Boost Interprocess, we can just change the version check so that it is compatible instead.

d235j:
The [release notes](http://www.boost.org/users/history/version_1_60_0.html) do not indicate any ABI changes.

cmake-local/BoostTargets.cmake
```diff
@@ -18,6 +18,9 @@ if(NEED_BOOST_LOCALE)
 endif()
 
 find_package(Boost 1.44 COMPONENTS ${required_boost_components} REQUIRED) # Lower version bound of 1.43 for range adapters, 1.44 for Filesystem v3
+if(Boost_VERSION GREATER 105900)
+    message(SEND_ERROR "Using an untested Boost version - inspect the Boost Interprocess release notes/changelog to see if any ABI breaks affect us.")
+endif(Boost_VERSION GREATER 105900)
```
rpavlik: For future reference, CMake hasn't needed repetition of the condition in the trailing statement for a long time (many many years). I'll just take care of it this time around, but just a heads-up.

rpavlik:
True, good point.  I'll just eyeball the diffs over in https://github.com/boostorg/interprocess since I don't always trust release notes. 

For instance, they changed the default `OffsetType` parameter of an `OffsetPtr` from `size_t` to `uintptr_t`, so I do actually have to look at our usage of that on Windows. Now, that might actually explain some 32-64 interop issues, but...
https://github.com/boostorg/interprocess/commit/e64e3b5acce023ce0da6e5c098ea52dd8a7b13e8

More review to go...

PR #324, opened 2016-01-12
Fix build race

rpavlik:
This will fix the intermittent failures related to the ReportTypesX header. We're now generating that, if required, at cmake time, rather than during the build, because I can't see how to make the dependencies work out right without making it a target of its own, which seems like overkill.


rpavlik:
@JeroMiya I'm assuming this was OK to merge, right?

PR #340, opened 2016-01-20
Idle server

rpavlik:
This pull request adds some CommonComponent functionality that might be suitable for future use by devices to idle themselves. (In a fake-out, I thought I'd be able to use it to idle the whole server, but I was mistaken.)

It also contains the code to idle the whole server - that is, if we have no clients connected (not counting internal/analysis clients), we set our sleep to 1 millisecond (ends up being more than that, but...) - incurring a bit of latency but saving a lot of CPU time. As soon as the first client connects, and until the last connection drops, our sleep goes back down to the configured value (which defaults to 0 - no sleep) for performance and low latency.

### DIFF
--- a/src/osvr/Common/CommonComponent.cpp
+++ b/src/osvr/Common/CommonComponent.cpp
@@ -61,36 +61,30 @@ namespace common {
     }
 
     void CommonComponent::registerPingHandler(Handler const &handler) {
-        if (m_pingHandlers.empty()) {
-            m_registerHandler(&CommonComponent::m_handlePing, this,
-                              ping.getMessageType());
-        }
-        m_pingHandlers.push_back(handler);
+        /// Just forward to the templated implementation.
+        registerHandler(ping, handler);
     }
 
     void CommonComponent::registerPongHandler(Handler const &handler) {
-        if (m_pongHandlers.empty()) {
-            m_registerHandler(&CommonComponent::m_handlePing, this,
-                              pong.getMessageType());
-        }
-        m_pongHandlers.push_back(handler);
+        /// Just forward to the templated implementation.
+        registerHandler(pong, handler);
     }
 
     CommonComponent::CommonComponent() {}
     void CommonComponent::m_parentSet() {
         m_getParent().registerMessageType(ping);
         m_getParent().registerMessageType(pong);
+        m_getParent().registerMessageType(gotFirstConnection);
+        m_getParent().registerMessageType(gotConnection);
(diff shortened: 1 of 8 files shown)


src/osvr/Common/CommonComponent.cpp
```diff
@@ -35,8 +35,24 @@ namespace osvr {
 namespace common {
     namespace messages {
 
+        // These messages all must match the ones in VRPN exactly.
+        /// @todo add tests to verify that these message identifiers match VRPN
+        /// string constants.
+
```
JeroMiya: Does this todo need to be done before merging this pull request?

src/osvr/Server/ServerImpl.cpp
```diff
@@ -101,9 +100,30 @@ namespace server {
...
+            vrpnConn->register_message_type(vrpn_got_first_connection),
+            &ServerImpl::m_exitIdle, this);
+        vrpnConn->register_handler(
+            vrpnConn->register_message_type(vrpn_dropped_last_connection),
+            &ServerImpl::m_enterIdle, this);
     }
 
-    ServerImpl::~ServerImpl() { stop(); }
+    ServerImpl::~ServerImpl() {
+#if 0
```
JeroMiya: Could you add a comment here explaining what this disabled code is for or why it's disabled?

JeroMiya:
Looks good to me on a sight read. Couple of minor comments (see above) but nothing worth blocking the merge over.

src/osvr/Common/CommonComponent.cpp
rpavlik: well, it's probably more likely to actually get done if it blocks the merge, if I'm being honest about it.

src/osvr/Server/ServerImpl.cpp
rpavlik: Typically handlers are unregistered - don't have easy access to the pointer or other things to unregister here,but it didn't seem to cause a problem. The commenting out was my abdication of a decision, probably a poor choice. I can stash a functor to unregister.

src/osvr/Common/CommonComponent.cpp
rpavlik: OK, adding tests for the new stuff. The existing stuff is harder to test, but test coverage isn't decreasing with this pull request once updated.

rpavlik:
OK, all tests passed locally.

rpavlik:
OK, that last little tweak was just for non-default configs. Still tests out just fine.

PR #341, opened 2016-01-20
Adjust required boost versions.

### LINKED ISSUE #323: Investigate Boost 1.60
IPC compatibility, specifically.  See http://www.boost.org/users/history/version_1_60_0.html

https://github.com/boostorg/interprocess/compare/boost-1.59.0...boost-1.60.0

related: https://github.com/OSVR/OSVR-Core/pull/320

rpavlik:
Accept boost 1.60 without error, warn-early on boost <1.54 on Windows,
and in the boost version warning, mention the file that will fail to compile
if you just comment out the version check.

Fixes #323.

It doesn't remove the error message from the C++ file, but it does warn you in the CMake build that if you're just going to comment out the error check, you're going to have to do it twice because it's there for a good reason.

### DIFF
--- a/cmake-local/BoostTargets.cmake
+++ b/cmake-local/BoostTargets.cmake
@@ -12,14 +12,22 @@ if(NEED_BOOST_FILESYSTEM)
     list(APPEND required_boost_components filesystem)
 endif()
 
-
 if(NEED_BOOST_LOCALE)
     list(APPEND required_boost_components locale)
 endif()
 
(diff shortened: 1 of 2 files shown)


rpavlik:
@d235j Does this seem like a reasonable compromise to you? I know you wanted the error pulled from the C++, but I feel more secure having it there, that it's somehow closer to the thing that would break if preconditions were violated. I'll merge it for now to make things easier.

d235j:
Yeah I think this is fine.

PR #342, opened 2016-01-22
Adding VRPN device descriptors

godbyk:


apps/external-devices/device-descriptors/5DT_Data_Glove_14_Utra.json
```diff
@@ -0,0 +1,13 @@
+{
+    "deviceVendor": "Fifth Dimension Technologies",
+    "deviceName": "5DT Data Glove 14 Utra",
```
rpavlik: spelling?

Also, I don't know exactly the mapping for the 14 ultras, but I think I know the joint mapping for the 5 ultras, not sure how important that is.

apps/external-devices/device-descriptors/Atmel_ATmega32.json
```diff
@@ -0,0 +1,13 @@
+{
+    "deviceVendor": "Atmel Corporation",
+    "deviceName": "ATmega32",
```
rpavlik: I'd add the caveat "with custom firmware" since the atmel driver doesn't just work with a bare unflashed avr.

apps/external-devices/device-descriptors/Oculus_Rift_DK2.json
```diff
@@ -0,0 +1,16 @@
+{
+    "deviceVendor": "Oculus VR",
+    "deviceName": "Rift DK2",
```
rpavlik: maybe identify that this is "Raw USB driver" (same for the dk1)

apps/external-devices/device-descriptors/5DT_Data_Glove_14_Utra.json
godbyk: @rpavlik Thanks! I'll fix this in the next batch.

Right now I'm skipping the mappings. The primary goal at the moment is to flesh out the supported devices table on the website.

apps/external-devices/device-descriptors/Atmel_ATmega32.json
godbyk: @rpavlik Can we add a notes field for things like this? A lot of devices have caveats associated with them (e.g., device only works in this particular mode or device has variable number of analog/digital inputs that must be use-specified).  It'd also be nice to have a url field that could point to the manufacturer's product page.

rpavlik: You can certainly add one, that's part of the reason we used JSON - we're not punishing unrecognized properties. I'd certainly suggest a notes field for these - probably standardizing a notes field to generate onto the web site, because for the external devices, where the server isn't built into the vendor's device, the user will want to know what the name of the device is for their `vrpn.cfg` file.

apps/external-devices/device-descriptors/5DT_Data_Glove_14_Utra.json
rpavlik: Yep - realized that after I posted that note.

PR #346, opened 2016-02-02
Change default config on win

rpavlik:
This makes the `osvr_server_config.json` file copied into the build tree and installed into the install tree by default on Windows platforms the HDK 1.3 direct mode config that supports positional tracking, etc. The original file named that in the source is still shipped, just under a different name causing it to not be the default.  On non-windows platforms there should be no change in behaviors.

Snuck in a little fix that should hopefully also fix Travis-CI's whining while I was at it.


rpavlik:
So it does what it says on Windows, and fixes the Linux Travis. Can someone merge this?

PR #348, opened 2016-02-05
Added error message asking the user if he has built the plugin for th…

peroht:
…e correct platform.

### DIFF
--- a/inc/osvr/Server/ConfigureServerFromFile.h
+++ b/inc/osvr/Server/ConfigureServerFromFile.h
@@ -113,6 +113,7 @@ namespace server {
                     out << " - " << pluginError.first << "\t"
                         << pluginError.second << endl;
                 }
+				out << "Is the plugin built for the desired platform?" << endl;
                 out << "\n";
             }
 


rpavlik:
I'm guessing that you were bit by this one?  Message is a good idea, though might tweak the wording to clarify how/what the issue might be. Something like "This is a 32 bit server, make sure the plugin you're using is built for 32 bit" or vice versa, since we can detect that at build time.

peroht:
Yes, it hit me when i now first tried understanding plugin development. What do you suggest? I have not thorough boost experience.

-------- Originalmeddelande --------
Från: Ryan Pavlik <notifications@github.com> 
Datum: 2016-02-06  02:51  (GMT+01:00) 
Till: OSVR/OSVR-Core <OSVR-Core@noreply.github.com> 
Kopia: Peter Thor <peter.thor@vicator.com> 
Rubrik: Re: [OSVR-Core] Added error message asking the user if he has built the plugin for th… (#348) 

I'm guessing that you were bit by this one?  Message is a good idea, though might tweak the wording to clarify how/what the issue might be. Something like "This is a 32 bit server, make sure the plugin you're using is built for 32 bit" or vice versa, since we can detect that at build time.

PR #360, opened 2016-02-18
Unified video tracker

rpavlik:
This was intended to replace the two-plugin solution, but performance isn't quite there yet. It's most likely a parameter tuning issue at this point, since I unit tested and found/fixed some bugs and it's acting nicer now. Does improve the existing tracker, too: could actually turn back on the right side beacons if we wanted to with minimal negative impact, I think.  Would like to get this merged so it doesn't drift away and so it can be picked back up at some later time.


rpavlik:
OK, this has been updated with some improvements:

- The original/standard video tracker has been improved by re-enabling the right-side beacons on the HDK: the tracking issue noticed that caused us to disable them was resolved when developing this branch.
- Most files/code that were identical between the two implementations are now in a separate directory shared between the two implementations, so there's only one copy of it, to avoid getting out of sync. Exceptions to this:
  - Image sources: In the new plugin, I made this into a proper library instead of just some files as in how I did it in the old plugin. I probably should pull the library from the new plugin into the shared dir,  to use it in both places, since this is a chunk of code, and contains platform-specific stuff, so we really don't want to needlessly duplicate this, lest someone port tracking to Linux twice.
  - LED identifiers: the new code uses typesafe IDs instead of bare ints, which is the only difference. It's a minor change for safety/reliability, but meant I couldn't just move files and delete files to make them share. It's probably worthwhile to backport this, however, since it's not a lot of effort. (Just taking your favorite diff tool - meld, winmerge, what-have-you, between the two directories, will help you do most of it nearly-automatically, might be a nice community task if someone is up to it.)

Ready to merge. Someone can go ahead and do it when Travis is happy.

PR #367, opened 2016-02-22
Plugin search path unique, single plugin instance

rpavlik:
Greg, Kevin, and I identified some issues with plugin loading - that we may have been loading plugins multiple times for a number of reasons (paths being present on the plugin search path multiple times, not checking to see if a plugin was already loaded by a particular name). This branch fixes those two bugs and should also provide a performance boost to server startup by only populating that search path once instead of 6-10 times (which was being made more expensive with the de-duplication added)


rpavlik:
cc @godbyk @DuFF14

rpavlik:
This might also be related to the long-standing "opens the hydra twice" issue. In any case, got two reviews, and confirmation that it fixed the issue that was being noticed with the new oculus plugin, so merging.

PR #368, opened 2016-02-23
Only build osvr_list_usbserial if BUILD_USBSERIALENUM is enabled.

godbyk:
The `osvr_list_usbserial` app fails to build if BUILD_USBSERIALENUM isn't being built and if it can't find the `osvr/USBSerial/USBSerialEnum.h` header file.


rpavlik:
Merged it into my cleanup branch, thanks!

PR #372, opened 2016-02-25
DK2 Display Descriptor with k1 and polynomial distortion.

DuFF14:


rpavlik:
Is this still an active item or can we close it? see also https://github.com/sensics/OSVR-RenderManager/issues/232

Issue #338, opened 2016-01-18
Needs updated HIDAPI VRPN


rpavlik:
This will presumably result in a few changes here, but for Linux (and Mac) support to work properly with much hardware, we will need to pull updated VRPN with updated HIDAPI (and/or, modify OSVR and VRPN to also be able to handle system-built HIDAPI, which would fix things on Linux I think and probably make it easier to package there).

The corresponding VRPN pull request is https://github.com/vrpn/vrpn/pull/130 - so the current workaround for a nicely-working OSVR Core on Linux is to just update the vendor/vrpn submodule to that pull request.

rpavlik:
Note that as a temporary workaround, you can build this branch: https://github.com/OSVR/OSVR-Core/tree/non-windows-workaround being sure to run `git submodule update` - its only difference from master as I type this is the state of the VRPN and HIDAPI submodules.

To clone this branch fresh, try

```bash
git clone --recursive https://github.com/OSVR/OSVR-Core.git --branch non-windows-workaround
```

agirault:
+1
Support for Linux and MacOSX is I believe a big asset of OSVR compared to other VR solutions.

rpavlik:
This is fixed and workaround no longer required! You can use the master branch since #373 was merged - just make sure that if you're using an existing git clone, that you run `git submodule update --recursive` after switching branches.

Issue #321, opened 2016-01-10
osvr_server crashes on launch on Mac OS X in hidapi code


d235j:
Currently, osvr_server crashes on Mac OS X with the following output:
```
[OSVR] Performing hardware auto-detection.
2016-01-08 13:49:23.233 osvr_server[45387:1885760] *** Terminating app due to uncaught exception 'NSRangeException', reason: '-[NSTaggedPointerString getCharacters:range:]: Range {0, 32} out of bounds; string length 3'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff907c3ae2 __exceptionPreprocess + 178
	1   libobjc.A.dylib                     0x00007fff86c7bf7e objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff907c398d +[NSException raise:format:] + 205
	3   CoreFoundation                      0x00007fff9075eaba -[NSTaggedPointerString getCharacters:range:] + 394
	4   CoreFoundation                      0x00007fff9068afc6 __CFStringEncodeByteStream + 4054
	5   com_osvr_Multiserver.so             0x000000010967b4fd make_path + 123
	6   com_osvr_Multiserver.so             0x000000010967b2c2 hid_enumerate + 408
	7   com_osvr_Multiserver.so             0x000000010966036f _ZN18VRPNHardwareDetectclEPv + 69
	8   com_osvr_Multiserver.so             0x00000001096602b4 _ZN4osvr4util19functor_trampolines6detail6CallerINS1_11this_last_tELi1ELi1EE11SpecializedIPFcPvS7_E18VRPNHardwareDetectE4callES7_S7_ + 18
	9   libosvrPluginHost.0.dylib           0x000000010929d47c _ZN4osvr10pluginhost37PluginSpecificRegistrationContextImpl30triggerHardwareDetectCallbacksEv + 32
	10  libosvrPluginHost.0.dylib           0x000000010929ef16 _ZN4osvr10pluginhost19RegistrationContext21triggerHardwareDetectEv + 28
	11  libosvrServer.0.dylib               0x00000001091d9ea0 _ZN4osvr6server10ServerImpl8m_updateEv + 214
	12  libosvrServer.0.dylib               0x00000001091da055 _ZN4osvr6server10ServerImpl6m_loopEv + 43
[...]

PR #373, opened 2016-02-27
Update VRPN and HIDAPI

rpavlik:
This pull request updates the VRPN submodule, and its HIDAPI submodule, to the latest upstream versions. This provides a single source tree that works across platforms, and improved compatibility.

It also fixes detection and initial use of Hydras on Linux.

Fixes at least #338 and #321


rpavlik:
OK, updated a few things in VRPN/HIDAPI noted by @janoc that I didn't see in local testing, that we'll chalk up to win8.1/win10 different but might be timezones or phases of the moon or something. Works well on 8.1 and 10 (and Linux) locally here now, so merging.

PR #375, opened 2016-03-01
Example tracker

JeroMiya:


rpavlik:
Do you want it as an autodetect plugin, or as a configured one? If it were a configured one, then you could have the simulated movement be optional.

examples/plugin/CMakeLists.txt
```diff
@@ -39,6 +40,7 @@ foreach(pluginname
     com_osvr_example_Configured
     com_osvr_example_EyeTracker
     com_osvr_example_Locomotion
+	com_osvr_example_Tracker
```
rpavlik: this doesn't line up on GitHub, suspect tabs/spaces mismatch. Try using Atom or Notepad2-mod with visible whitespace.

examples/plugin/com_osvr_example_Tracker.cpp
```diff
@@ -0,0 +1,122 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Internal Includes
+#include <osvr/PluginKit/PluginKit.h>
+#include <osvr/PluginKit/TrackerInterfaceC.h>
+
+// Generated JSON header file
+#include "com_osvr_example_Tracker_json.h"
```
rpavlik: Recommending all new plugins be named `org_osvr_whatever` - and if this is intended as a dummy plugin, not just an example (since there are examples that put out tracker data, aren't there?), maybe namespace that name differently.

examples/plugin/CMakeLists.txt
```diff
@@ -5,7 +5,8 @@ set(OSVR_EXAMPLE_DEVICE_PLUGINS_SIMPLE
     com_osvr_example_DummyDetectAndCreateAsync
     com_osvr_example_EyeTracker
     com_osvr_example_Locomotion
-    com_osvr_example_MultipleAsync)
+    com_osvr_example_MultipleAsync
+    com_osvr_example_Tracker)
```
rpavlik: So by building the plugin here, as opposed to separately down below, you're accepting the manual-load, no-install properties of all those other plugins. Just wanted to make sure that was intentional.

examples/plugin/com_osvr_example_Tracker.cpp
```diff
@@ -0,0 +1,122 @@
+/** @date 2015
```
rpavlik: I think someone copied and pasted... unless you started this three months ago...

examples/plugin/com_osvr_example_Tracker.cpp
```diff
@@ -0,0 +1,122 @@
...
+        m_dev.initAsync(ctx, "Tracker", opts);
+
+        /// Send JSON descriptor
+        m_dev.sendJsonDescriptor(com_osvr_example_Tracker_json);
+
+        /// Register update callback
+        m_dev.registerUpdateCallback(this);
+
+        /// Seed pseudo random generator.
+        std::srand(std::time(0));
```
rpavlik: You're not actually using any old c random stuff later on, as far as I can tell, so this line (and a bunch of includes), can be dropped.

examples/plugin/com_osvr_example_Tracker.cpp
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    OSVR_ReturnCode update() {
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(
+            5)); // Simulate waiting a quarter second for data.
+
+        OSVR_TimeValue now;
+        osvrTimeValueGetNow(&now);
+        double t = 5.0 * ((double)now.seconds + ((double)now.microseconds / 1000000.0));
```
rpavlik: I don't understand the purpose of this line: it looks like it's trying to get the time since epoch in seconds, but then multiply it by 5.  If you want time since startup, you can either use `std::chrono` stuff (some of this would go in the class, some in the constructor, some in the update) or `OSVR_TimeValue`, but I'd recommend `chrono`

```c++
using our_clock = std::chrono::steady_clock; // like a typedef, stick it somewhere useful.

our_clock::time_point m_startTime; // class member
m_startTime(our_clock::now()) // in the constructor initializers

// then in update,
std::chrono::duration<double> secondsSinceStart = our_clock::now() - m_startTime;
double t = secondsSinceStart.count(); // doesn't do the * 5
```

then if you want a function of `sin(5 * t)`, do it like that, because `t` sounds like time, not time with a scale factor. (and yes, I understand that this probably originated in a similar simulated/dummy data plugin of another kind - I just didn't get asked for a review on it and/or didn't do a good enough one on it)

examples/plugin/com_osvr_example_Tracker.cpp
```diff
@@ -0,0 +1,122 @@
...
+class HardwareDetection {
+  public:
+    HardwareDetection() : m_found(false) {}
+    OSVR_ReturnCode operator()(OSVR_PluginRegContext ctx) {
+
+        if (m_found) {
+            return OSVR_RETURN_SUCCESS;
+        }
+
+        std::cout << "PLUGIN: Got a hardware detection request" << std::endl;
```
rpavlik: maybe remove the printouts?

examples/plugin/com_osvr_example_Tracker.cpp
JeroMiya: Time since epoch vs time since start doesn't matter - the values I'm sending for position are cyclical. The point was just to get some kind of movement in there so you can tell it's working. t still represents time, just in some unnamed unit defined as 5 units per second.

PR #379, opened 2016-03-08
Build fixes

rpavlik:
Should be widely safe. (and required, for msvc2015, and possibly other platforms - the findjsoncpp that's in there is goofed up, and I've heard others hitting the OpenCV issue I fixed, etc.)


godbyk:
:+1:

JeroMiya:
Does this fix the issue I've been having where it would lose jsoncpp whenever it re-ran cmake during a build?

rpavlik:
Yes, it fixes that.

PR #381, opened 2016-03-08
Json load enhancements

rpavlik:
The primary benefit here is better error messages if you can't load your config/display descriptor/render manager config - since it will distinguish "can't open file" from "can't parse file"


JeroMiya:
Looks good from a sight read. This will make at least one support ticket easier for me to diagnose! :)

PR #382, opened 2016-03-09
Adjust server and path tree to pass port more often.

rpavlik:
Includes updates required to JointClientKit, which doesn't use a port, and thus needs to signal that we shouldn't add a port to path tree entries for it.

Also splits the methods for adding a device node from an "external devices" entry from the method used to add a device node from a new device created by OSVR, so they may have different default ports.

These are in preparation for changing the port, but do not break any public API, ABI, or network interface.


rpavlik:
To transport some messages from office discussion: this is mostly a "plumbing" change - doesn't affect even most of the work devs on the innards do (mostly hidden in the server implementation factories). It reduces the amount we're relying on VRPN to infer the port we want to use, in favor of explicitly specifying our intended port in the path tree when possible. This is in anticipation that once OSVR moves to its assigned port, VRPN, as a transport, will infer the wrong port for OSVR-hosted devices.

PR #383, opened 2016-03-09
Fix jsoncpp

rpavlik:
This should fix the JsonCpp issues we'd been having since an earlier improvement to the cmake script. Tested on Windows with new and old, 32 and 64 bit libraries, in both shared and static builds, as well as on Linux with both distribution-built (so, ancient and without cmake config file) and newer builds.


rpavlik:
Got a few verifications that this does fix things.

PR #388, opened 2016-03-10
Set numBuffers to 2 and enable vsync

DuFF14:
To match new RenderManager install sample configs. @russell-taylor should vsyncenabled be true or false in the non-direct mode configs?


rpavlik:
I don't believe the non-directmode path respects that config variable. In any case, your committer name/email was not set when you committed this.

DuFF14:
Thought I had fixed this but still looks like an anon commit. Opened a new PR https://github.com/OSVR/OSVR-Core/pull/410

PR #391, opened 2016-03-14
Vendor asio

rpavlik:
Adds a new vendored-in library, ASIO - http://think-async.com/Asio/Documentation - that can be used for cross-platform networking, as well as asynchronous or synchronous serial port and other IO operations. It's a C++ standards-track library that has a good track record and adds no new dependencies, not even additional usage of Boost in its latest version (and it's header-only).

Is used in a forthcoming (1.0) change that will watch an old port and warn if anyone tries to connect to it, but I wanted to get this in and available before 1.0.

This also includes a small patch to the clang-format style so that if you update your version of clang-format you don't start mysteriously sorting includes.


russell-taylor:
I worry about library creep.  Both because potential users can be (incorrectly, but nonetheless) turned off when they see how many other libraries OSVR requires them to pull in and because of experience with incompatible versions and trying to track library changes.  How can it take a new library to do something we were already doing?

godbyk:
How does this version of asio differ from the implementation in boost or the proposed implementation in the standard C++ library?

rpavlik:
The version in boost is automatically generated from this version at boost release time. As far as I can tell, this is the closest code to the proposed standard.

Re: library creep, it's a vendored in header only library, so most users won't notice. If by "we were doing before" you mean in VRPN, yes, there's socket code in there, but it's not factored out to be easily reusable. Additionally, asio uses (optionally, if available on the platform at runtime) faster platform specific backend code like io completion ports on Windows.  (I was actually considering seeing if I could use the existing implementation hiding in VRPN to allow an optional asio socket implementation, similar to your optional chrono time backend, and see what that did to performance and or the handshake issue)

rpavlik:
(There's literally a script called boostify in the source tree. I could have used the version in boost but given the anti boost bias out there, as well as the fact that the non boost version can be vendored easier and updated faster than boost release cycles, it seemed wise to go this route instead)

PR #393, opened 2016-03-15
Add hand-tuned distortion coefficients on the HDK 1.1/1.2

mdutton3:
Modified HDK 1.1 display config with hand-tuned distortion coefficients. Ignores chromatic aberration.

Should the we provide two display configs? One with and one without pre-distortion?

### DIFF
--- a/apps/displays/OSVR_HDK_1_1.json
+++ b/apps/displays/OSVR_HDK_1_1.json
@@ -26,9 +26,12 @@
       }
     ],
     "distortion": {
-      "k1_red": 0,
-      "k1_green": 0,
-      "k1_blue": 0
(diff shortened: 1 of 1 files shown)


rpavlik:
Sorry it took so long to get to this - I do like the idea of providing an optional config with distortion. (That said, I think in RenderManager, even devices with no distortion listed basically just get a mesh defined by an identity polynomial, so I don' tthink there's a performance impact there). Will try this out on my 1.2.

russell-taylor:
Indeed, RenderManager always does 3-color distortion, which is the identity in each case by default.

rpavlik:
Sorry it took so long, but just tested this, looks nice!

PR #395, opened 2016-03-15
Client side prediction

russell-taylor:
Adds ability to specify client-side prediction in RenderManager config.


russell-taylor:
Continues to work with old config files, so going ahead and merging into master.  This will make it easier to combine with what will hopefully soon be a fix for transforming angular velocities.

PR #398, opened 2016-03-21
Implements transformations for angular velocity quaternions correctly.

russell-taylor:


rpavlik:
Thanks for this! the logic makes sense - I'm working on pulling it in, removing workarounds, etc. now.

rpavlik:
So my only test mechanism here is using RenderManager and adjusting the client-side prediction interval to something big-ish and trying to figure out if what I see is what I expect.

I've put all my config files for doing this here: https://gist.github.com/rpavlik/fc743b6a2dd655fce47f75c30fcba955

My two cases are: 
- running with `osvr_server_config.basic.clientside.json` - directly using the IMU/prediction
- running with `osvr_server_config.clientside.json` (but with no IR cam plugged in) - this is close to the default config (uses the IMU fusion plugin) which takes the velocity apart, puts it back together, and now, sends it along (and all workarounds have been removed from that path).

(the render manager config goes into `sample-configs` to run this)

Ideally, they'd both behave the same, and they'd both behave as expected - which I think is overrotating in the direction of rotation. That's not what appears to be happening - on my unit, the first (simplest) case, I appear to get some pitching what I try to just yaw...

Would appreciate a second look, and/or a correction of my expectations before merging.

russell-taylor:
This works for me in an orientation-only test using client-side prediction on RenderManager with an HDK 1.3 when I sit the HDK on a table top; when I rotate in one axis at a time, I get overcompensation in one eye for my config file that has one eye predict 500ms into the future and the other 16ms.  Did not have the camera plugged in or the fusion tracker running.

This also works for me when I disable client-side prediction and use the info from the DeadReckoningRotationFilter.sample.json to increase the prediction on the server side.  I get both eyes predicting ahead in my one-axis-at-a-time test.

So I think this is working.

PR #399, opened 2016-03-21
Adding client-side prediction entries to the DirectMode and non-Direc…

russell-taylor:
…tMode RenderManager configs assuming that they are being applied to an HDK.

Varies the prediction interval for landscape/portait and direct/nondirect.
Adds expected HDK sensor delay to all of them.

### DIFF
--- a/apps/sample-configs/renderManager.direct.landscape.json
+++ b/apps/sample-configs/renderManager.direct.landscape.json
@@ -23,6 +23,14 @@
             "bitsPerColor": 8
         },
 
+		"prediction": {
+				"enabled": true,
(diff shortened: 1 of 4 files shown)


rpavlik:
Do the prediction amounts in here account for the fact that there's already prediction being done server-side? (I ask since you mention "expected HDK sensor delay") Also, how will this work with the positional tracker once I get that reporting velocities too?

russell-taylor:
The prediction amounts assume no server-side prediction.  The sensor delay is added to the common prediction interval, which is used for both eyes.  Positional prediction already in the code but not yet tested in combination with orientation.

Issue #155, opened 2015-06-18
Interface for plugins to trigger hardware detect


rpavlik:
Sometimes SDKs can tell us when a new device shows up. While we could probably just add the device then, we might get into threading issues depending on a number of factors. An easy way of handling this would be to be able to tell the server from the plugin, "please hardware detect when you get a chance". Related to #153 in this way.

rpavlik:
And this basically is just wiring the API parts, since there is an internal API for triggering hardware detect of course.

PR #404, opened 2016-03-24
Added API to PluginKit so plugins can trigger hardware detection.

godbyk:
This PR addresses issue #155.


rpavlik:
Also needs rebasing by now (sorry) and possibly analysis/commenting of thread safety - may need lock guard in case it's called from an async plugin, etc.

godbyk:
I rebranched from master and applied this PR patch again so it should merge cleanly now.

As for thread safety, I think we're okay there because we just set a trigger-hardware-detect flag to true (in `ServerImpl`) and the triggering is actually executed the next time the server loops through.

PR #410, opened 2016-03-30
Changed default numBuffers to 2 with vsync enabled.

DuFF14:


DuFF14:
Can we merge this? Will also want to update when ATW is working.
How about client prediction, are we leaving that out for now?

DuFF14:
@russell-taylor are the settings in these RenderManager config files what we want to ship by default with the OSVR installation?

DuFF14:
Should prediction be in there by default? Is that working with all client apps?

russell-taylor:
Prediction should be there by default, and should work with all client apps.  I don't know that we've yet settled on the best all-around render params (1 vs. 2 buffer, amount of wait, etc.).

The only issue with turning on client-side prediction is that it doesn't work with servers that don't send velocity info, and the current predicting server does not.  On the other hand, it does no harm -- the client side just avoids predicting in that case.

DuFF14:
Updated OSVR_HDK_1_3_with_mesh to use built-in distortion mesh v2 and new FOV values.
Updated default renderManagerConfig settings -- enabling/disabling vsync settings for direct/extended modes, decreased renderOverfillFactor from 1.5 to 1.2.

I didn't change the OSVR_HDK_1_3 COP or FOV values, as it didn't look correct with the polynomial coefficient distortion method. But it seems odd to have different FOV values for the same display...

russell-taylor:
The FOV depends on the distortion correction.  Two different corrections can have different FOVs so long as they move the pixels into consistent places in their distortion correction.

rpavlik:
Is this outdated and closeable, or still an active issue?

PR #418, opened 2016-04-15
Android autostart

JeroMiya:
This is an implementation of the auto-start API for Android. It requires linking osvrClientKit to osvrServer, but this is only done for the Android platform.

Requires this libfunctionality PR to be merged: https://github.com/OSVR/libfunctionality/pull/15

### DIFF
--- a/src/osvr/ClientKit/ServerAutoStartC.cpp
+++ b/src/osvr/ClientKit/ServerAutoStartC.cpp
@@ -28,7 +28,9 @@
 #include <osvr/Util/PlatformConfig.h>
 #include <osvr/Client/LocateServer.h>
 #include <osvr/Common/GetEnvironmentVariable.h>
+#if defined(OSVR_ANDROID)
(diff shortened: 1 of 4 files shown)


rpavlik:
I think this looks good, as long as the various consumers of the Android port (Unity CI? others?) are sure to verify that they're including the sufficient dependency libraries (basically that clientkit implies bring the whole family).

src/osvr/ClientKit/ServerAutoStartC.cpp
```diff
@@ -38,6 +38,7 @@
 // Standard includes
 // - none
 
+// @todo use a thread-safe lazy-initialized singleton pattern
 #if defined(OSVR_ANDROID)
 static osvr::server::ServerPtr gServer;
```
rpavlik: Did you want to do anything about this before merge, or will you file an issue, or is it not expected to be a big deal?

PR #423, opened 2016-04-21
Thread-safety for device descriptor sending in async plugins.

rpavlik:
I think I hit this in early revisions of the vive plugin - it seems like the original implementation of the device descriptor sending code assumed you'd either only send it once, or you'd be a sync plugin and send from the main thread: there was no send-guard or equiv. locking around the descriptor update but that's definitely not something you want to do in a non-main thread since it can trigger a send.

This hasn't been tested, don't merge it until it has been.

### DIFF
--- a/src/osvr/Connection/DeviceToken.cpp
+++ b/src/osvr/Connection/DeviceToken.cpp
@@ -116,6 +116,8 @@ bool OSVR_DeviceTokenObject::releaseObject(void *obj) {
 
 void OSVR_DeviceTokenObject::setDeviceDescriptor(
     std::string const &jsonString) {
+    auto guard = getSendGuard();
+    guard->lock();
     m_getConnectionDevice()->setDeviceDescriptor(jsonString);
     m_getConnection()->triggerDescriptorHandlers();
 }


src/osvr/Connection/DeviceToken.cpp
```diff
@@ -116,6 +116,8 @@ bool OSVR_DeviceTokenObject::releaseObject(void *obj) {
 
 void OSVR_DeviceTokenObject::setDeviceDescriptor(
     std::string const &jsonString) {
+    auto guard = getSendGuard();
+    guard->lock();
```
godbyk: Do we need to call `unlock()` or does a destructor take care of that? (I gave up tracing through the code after 10 minutes.)

rpavlik:
It's a guard class - so raii handles the unlock. It just needs a manual lock because it's polymorphic and the actual guard, if applicable, is returned, not a lockable object. (Tracing it shouldn't be that hard, think it's just guard interface in util).

godbyk:
The `GuardInterface` destructor was empty, which is why I asked.

rpavlik:
Well, right, the base class is, since sync devices don't need to pay
locking costs. If you have an async device token, your send guard should be
derived from guard interface but actually do something, iirc.

On Sat, Apr 23, 2016, 3:08 PM Kevin Godby <notifications@github.com> wrote:

> The GuardInterface destructor was empty, which is why I asked.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/OSVR/OSVR-Core/pull/423#issuecomment-213821713>
>
-- 
Ryan A. Pavlik, Ph.D.
Senior Software Engineer
Sensics, Inc.
www.sensics.com

PR #424, opened 2016-04-22
Timer frequency

rpavlik:
Apps are probably bumping the timer frequency already, but in case they aren't, we don't want to accidentally miss milliseconds.  This patch adds an RAII class to increase/decrease the timer frequency, and hooks it up to the server idle mechanism (so we won't spin up the timer frequency if we're supposed to be idling).

 Even @randomascii says it's OK, because VR is a special case where you can see milliseconds :sweat_smile: https://randomascii.wordpress.com/2016/03/08/power-wastage-on-an-idle-laptop/#comment-20184


randomascii:
:-)

Someday maybe Microsoft will give us a better way of doing accurate waits. Until then, yep, sometimes necessary.

PR #425, opened 2016-04-22
Inverting the headSpace transform so that it ends up pushing head spa…

russell-taylor:
…ce forward.

This pushes the eyes forward of the origin, which makes sense because they are ahead of the center of rotation of the head.
This is consistent in TrackerViewer and in RenderManager with what is expected.

### DIFF
--- a/apps/sample-configs/osvr_server_config.HDK12DirectMode.sample.json
+++ b/apps/sample-configs/osvr_server_config.HDK12DirectMode.sample.json
@@ -26,7 +26,7 @@
     }],
     "aliases": {
         "/headSpace": {
-            "translate": [0.0, 0.0, 0.04141],
+            "translate": [0.0, 0.0, -0.04141],
             "child": "/org_osvr_filter_videoimufusion/HeadFusion/semantic/fused"
         },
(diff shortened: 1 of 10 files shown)


russell-taylor:
Actually, this distance is more like 0.06 for me, my eyes are around 6cm ahead of the center of my head rather than 4cm...

rpavlik:
the value you refer to is actually the distance between the origin of the video tracker body coordinate system and eye space, hence its size and direction. It's perhaps poorly named, and poorly located (it should be in the device descriptor or deeper in the system, not in the config file), but that's what it's actually for, not to provide a neck model.

russell-taylor:
I see.  Adding this offset to cases with camera tracking makes good sense.  However, in cases where there is not camera tracking and we're reporting only orientation any origin is as good as any other and this offset actually impedes the specification of a correct head model (because you need to back it out again later).  Even when not specifying a head model, having it in there causes objects in the world to move in the OPPOSITE direction from expected.  I recommend removing this from the case where there is not camera tracking.  I'll open a separate issue on adding a head model.

PR #431, opened 2016-05-02
Update com_osvr_Multiserver.cpp

Olm-e:
correct compiling error with additional parameters


rpavlik:
It looks like you are possibly using an older version of the VRPN submodule: are you on the master branch and have run git submodule update?  I added those parameters to the corresponding constructor in VRPN over a month ago by now probably...

rpavlik:
The additional parameters are required for building with the bundled VRPN. If you're having trouble building, you may have an outdated submodule or an older version of VRPN somewhere earlier in your include paths.

PR #440, opened 2016-06-15
Sort includes and minor cosmetic changes spotted by clang-3.9.

araujobsd:
Minor style changes and sort includes. Not all includes were sorted at this time.

Submitted by: Marcelo Araujo <araujo@FreeBSD.org>

### DIFF
--- a/src/osvr/AnalysisPluginKit/AnalysisPluginKitC.cpp
+++ b/src/osvr/AnalysisPluginKit/AnalysisPluginKitC.cpp
@@ -23,15 +23,15 @@
 // limitations under the License.
 
 // Internal Includes
+#include "../Connection/VrpnConnectionKind.h" /// @todo Internal header!
 #include <osvr/AnalysisPluginKit/AnalysisPluginKitC.h>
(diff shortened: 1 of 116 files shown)


araujobsd:
Seems Travis is broken by itself: The command "curl -sSL "http://llvm.org/apt/llvm-snapshot.gpg.key" | sudo -E apt-key add -" failed and exited with 2 during .

rpavlik:
Doing this all in one chunk is sometimes the way to go, but it's also prone to breakage and to breaking other pull requests. We've got too many pull-requests in-flight right now that merging this would make it hard to merge those. Instead, we can choose to sort includes as we want to file-by-file as we touch files, handling conflicts one at a time rather than the whole codebase at once. (I think at least one PR has the "sort includes" flag turned off in the clang-format config, and we'll probably be adjusting the max line length, etc...)

To make it easier for your contributions to fit in, you might try using `git clang-format` which just reformats the files that you've edited - keeping it from formatting things you haven't touched and making the diff huge.

PR #449, opened 2016-07-28
Work around missing move c'tor in some ostream implementations.

godbyk:
This was fixed in GCC 5 (see [this bug report](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54316)).


rpavlik:
Well, if you actually allocated an ostringstream anywhere in that code it might work ;) I was going to just have the move c'tor construct the new ostringstream from the old one's string - not ideal but workable. I think this is probably the better approach, though, thoughts?

PR #452, opened 2016-08-01
Commit: 948259e3244d9e39387a2c0734cc990aa833469c breaks cmake-3.2.0 that used to works fine.

araujobsd:
There are two ways to solve this problem, copy the missing cmake modules or use this approach that is way simple.

Error:
    /tmp/osvr/OSVR-Core/cmake/cmake-3.2.0-modules/FindPackageHandleStandardArgs.cmake
Call Stack (most recent call first):
  cmake/GetGitRevisionDescription.cmake:85 (find_package)
  cmake/GetGitRevisionDescription.cmake:128 (git_describe)
  vendor/vrpn/ParseVersion.cmake:36 (git_get_exact_tag)
  vendor/vrpn/CMakeLists.txt:82 (include)

### DIFF
--- a/cmake/cmake-3.2.0-modules/FindGit.cmake
+++ b/cmake/cmake-3.2.0-modules/FindGit.cmake
@@ -100,7 +100,7 @@ endif()
 # Handle the QUIETLY and REQUIRED arguments and set GIT_FOUND to TRUE if
 # all listed variables are TRUE
 
-include(${CMAKE_CURRENT_LIST_DIR}/FindPackageHandleStandardArgs.cmake)
+include(${CMAKE_ROOT}/Modules/FindPackageHandleStandardArgs.cmake)
 find_package_handle_standard_args(Git
                                   REQUIRED_VARS GIT_EXECUTABLE
                                   VERSION_VAR GIT_VERSION_STRING)


rpavlik:
Actually this can be simplified further - don't specify an absolute path at all. That module should always be found using the search path. The stock modules in CMake use an absolute path so that they find their version matched peers, which is how I imported this bug, good catch.

araujobsd:
Fixed at: https://github.com/OSVR/OSVR-Core/commit/9e191386c76008f3fb2e02b30eadb91e70e4ecf9

Thanks!

rpavlik:
Wait, why didn't the normal pull request process continue here?

rpavlik:
OK, I'm guessing you probably haven't done a lot with CMake before, and that's OK. This regression wasn't breaking any of the main CIs (neither the deployment CI nor Travis), and the workaround is just to upgrade past cmake 3.2.

I was hoping to see a revised version of the patch that I could merge from this pull request and give you credit for it. I was confused when I saw a patch come in from @osvr-phabricator (we have a phabricator instance? I don't have access, which is strange for a primary contributor and maintainer...) that hadn't been reviewed but had my name on it as a stamp of approval. Since it was just a one-line fix, I fixed it for you (see dc530dc03c26c1d80ba9342b178ad7d2f846ce6e ) instead of just reverting and re-opening the pull request - easier all around (and I could write CMake code in my sleep) though it doesn't give you as much credit, sorry.

Looking forward to future contributions, though it's OK if you just report issues without fixing them if they're not in your wheelhouse!

osvr-phabricator:
OK, cool that you fixed it in a proper way! Particularly I was not looking for credit but instead fix my issue with that cmake version.
Sorry to by pass the pull request process, I will be more careful next time. 

For the phabricator instance, I have my own non-public one for code review and so on, can't give you access there. But if you are interested to use Phabricator for code review, we can talk more.

Best,

PR #459, opened 2016-08-13
Marking SDL2 as required because the library fails to build and link …

russell-taylor:
…without it, at least on Visual Studio 2015 with OpenCV turned off.


rpavlik:
What library fails to build without it? It builds just fine on linux and mac, and I build it just fine without it on VC2015.3.

russell-taylor:
Okay, so now it is working - must have been because my computer's clock had just reset its time and everything was building strangely.  Sorry for the false alarm.

PR #460, opened 2016-08-16
Update Building.md

russell-taylor:


rpavlik:
Really? the system provided jsoncpp still doesn't work? I spent like a week on the findjsoncpp trying to get it just right so that even ancient jsoncpp shipped with distros worked fine so people didn't have to do this nonsense anymore...

russell-taylor:
Really, at least when I tried it.

PR #467, opened 2016-08-29
Update logging

godbyk:
This pull request updates spdlog to fix a few bugs:

  * Exceptions will only be thrown on construction and not during log calls.
  * Reduced the asynchronous log message queue size to reduce memory usage.
  * Adjusted log level conversion function to handle reduced selection of spdlog log levels.
  * The async msg queue gets flushed on exit now.
  * LineLogger replaced with (slightly modified) StreamProxy.

Note: In the upstream spdlog repository, there is a bug in the Android sink (it uses some extinct log levels). I've [patched that](https://github.com/OSVR/spdlog/commit/4c10727da74de507c60a70476ebb8f71c9c0175e), but haven't merged it in yet since I haven't tested it. If someone tests that spdlog branch, please let me know if it works okay (or even if you review the code and it looks okay...).


JeroMiya:
Could you add this change with this pull request? This will allow it to build on android:
image

godbyk:
@JeroMiya Done!

src/osvr/Util/LogRegistry.cpp
```diff
@@ -87,22 +87,39 @@ namespace util {
             auto spd_logger = spdlog::get(logger_name);
```
rpavlik: does this need to be in a `try`?

src/osvr/Util/LogRegistry.cpp
```diff
@@ -135,7 +152,8 @@ namespace util {
             spdlog::set_pattern(DEFAULT_PATTERN);
```
rpavlik: Do these need to be in `try`?

src/osvr/Util/LogRegistry.cpp
```diff
@@ -187,7 +205,7 @@ namespace util {
             // File sink - rotates daily
             std::string logDir;
             try {
-                size_t q_size = 1048576; // queue size must be power of 2
+                size_t q_size = 65536; // queue size must be power of 2
```
rpavlik: what I couldn't figure out, was what the units here are - characters? lines? We shouldn't be logging that much data that we need a massive queue.

src/osvr/Util/Logger.cpp
```diff
@@ -64,24 +65,34 @@ namespace util {
...
+            try {
+                auto console_logger = spdlog::stderr_logger_mt(name, false);
+                return makeLogger(name, console_logger);
+            } catch (...) {
+                std::cerr << "Failed to create a console logger to use as a "
+                             "fallback. Logging will be disabled entirely."
+                          << std::endl;
+                auto null_sink = std::make_shared<spdlog::sinks::null_sink_st>();
+                auto null_logger = std::make_shared<spdlog::logger>(name, null_sink);
+                return makeLogger(name, null_logger);
```
rpavlik: so if spdlog totally fails for some reason, we get no output past this point - is that ok? The previous fallback did actually fall back to plain old cerr IIRC.

src/osvr/Util/Logger.cpp
```diff
@@ -131,95 +139,90 @@ namespace util {
...
+            proxy << msg;
+            return proxy;
         }
 
-        detail::LineLogger Logger::debug(const char *msg) {
-            return logger_ ? detail::LineLogger{logger_->debug(msg)}
-                           : detail::LineLogger{name_, LogLevel::debug, msg};
+        Logger::StreamProxy Logger::debug(const char *msg) {
+            StreamProxy proxy(*this, LogLevel::trace);
+            proxy << msg;
```
rpavlik: I'd add another constructor to streamproxy to simplify this to a one-liner and make it less likely that the StreamProxy has to be moved (vs RVO).

src/osvr/Util/Logger.cpp
```diff
@@ -131,95 +139,90 @@ namespace util {
...
+            return proxy;
         }
 
         // logger.info() << ".." call  style
-        detail::LineLogger Logger::trace() {
-            return logger_ ? detail::LineLogger{logger_->trace()}
-                           : detail::LineLogger{name_, LogLevel::trace};
+        Logger::StreamProxy Logger::trace() {
+            return { *this, LogLevel::trace };
```
rpavlik: Is this somehow an implicit conversion? Does this save us a move-construction? The syntax looks weird, and it's not just because you didn't run clang-format here :-P

src/osvr/Util/Logger.cpp
```diff
@@ -263,12 +266,12 @@ namespace util {
...
-                logger_->flush();
-            } else {
-                std::cout << std::flush;
-                std::cerr << std::flush;
-            }
+            logger_->flush();
+        }
+
+        void Logger::write(LogLevel level, const char* msg)
+        {
```
rpavlik: We could regain the "absolute fallback" of `if (!logger_)` by just reverting the flush changes above and putting a conditional in this one function, now that we have all the stream handling entirely within our own codebase.

src/osvr/Util/LogRegistry.cpp
godbyk: @rpavlik I'm not positive, but my reading of the code suggests it might be the number of messages that can be queued up at one time. I think we could probably drastically reduce this value, but didn't want to be too rash until we've gained a little more experience with it. I'm not sure how full the queue gets in practice.

rpavlik: I'd err on the side of being more synchronous than less, since we were originally very synchronous (every OSVR_DEV_VERBOSE had an endl with its implied flush) and logging is typically most important when things are working badly.

src/osvr/Util/Logger.cpp
godbyk: @rpavlik The thing that bugged me about how you had coded the fallback previously were the proliferation of if..then checks against the loggers. I'd rather have a known-good fallback logger that we can unconditionally use if there was an error constructing a real logger. If we can devise a way to guarantee this, I'd be quite happy.

src/osvr/Util/Logger.cpp
godbyk: How dare you question my magick! ;-) It's just calling [this `StreamProxy` constructor](https://github.com/OSVR/OSVR-Core/blob/update-spdlog/inc/osvr/Util/Logger.h#L129) and returning the constructed object.

src/osvr/Util/LogRegistry.cpp
godbyk: @rpavlik It only makes a copy of a `shared_ptr`. If the copy operation throws, then we can't save you anyway (because copying a fallback `shared_ptr` would also likely throw). I figure it's best to let any OOM errors just propagate up the stack.

src/osvr/Util/LogRegistry.cpp
godbyk: I think `set_pattern()` only throws in OOM cases. `set_level()` is just a bunch of assignments of the new level to replace the old level.

src/osvr/Util/LogRegistry.cpp
godbyk: We could go _completely_ synchronous if you want. But if people use logging more frequently (e.g., as I do at higher verbosity levels), then it may cause lag.

The queue does get flushed on exit now. (Though I think we need to explicitly call `flush()` in any signal handlers we've established because they bypass the destructors.) Further, it messages of a specified log level and above should trigger an immediate flush.

src/osvr/Util/Logger.cpp
godbyk: That's a good idea. This comment also drew my attention to a horrible series of copy-paste errors.

src/osvr/Util/LogRegistry.cpp
russell-taylor: It is critically important that the system not block rendering when flushing non-error messages to the screen -- it is okay to have a separate logging-consumer thread that blocks on the disk write after gobbling up all queued client-thread messages, but having the display burp every time we (for example) report update rate is not acceptable.

Almost as important is making the error messages appear when the program crashes.  I'd be willing to have error messages cause a rendering glitch in order to have this happen.  That is, asynchronous on info/warning and sync on error.

If things really flush correctly at crash, then all async should be okay.

One thing to consider is whether the OS keeps up with the background saving and never causes the buffer to fill.  If it fills and blocks the client thread to dump all of that to disk, we're going to get a large render blip.  Again, okay if that is a different thread but not okay if the render hangs due to logging.

src/osvr/Util/LogRegistry.cpp
russell-taylor: Rule #0 of library design: The library shall not cause the program to terminate.  No matter how bad things get, no matter that we think there's nothing the app can do about the situation, we return an error code rather than exiting or throwing an exception.

rpavlik: Well, and my absolute fallback didn't use any shared pointers, and we should be able to build most of this to use unique_ptr anyway. Unless you are specifying a nothrow copy explicitly, we should be catching here.

src/osvr/Util/Logger.cpp
rpavlik: yeah, I didn't like it either, but I wanted to never fail, and the way we were basically just wrapping spdlog classes meant there needed to be checks all over. Fortunately your changes made it easier to have many fewer tests and still have an absolute fallback.

rpavlik:
@godbyk would still like to see the changes I asked about - the merge was more to "stop the bleeding" given how crash-prone the existing implementation was.

godbyk:
@rpavlik Yep. I'll submit a separate pull request with some more patches soon.

PR #468, opened 2016-09-02
update vrpn submodule

ChristophHaag:
On recent glibc union wait was removed and this commit is needed:
https://github.com/vrpn/vrpn/commit/70d47fec7e3be85f5f05c8ac5be99dacbd8be257


ChristophHaag:
So on appveyor the build system choked and on travis clang segfaulted while linking. Great. The change itself shouldn't really be a problem though. On Linux it compiles fine and a vrpn update shouldn't really break other platforms when one works.

godbyk:
@ChristophHaag It's safe to ignore the AppVeyor complaints. I haven't finished configuring the AppVeyor build yet.

The Travis CI builds succeeded but for one instance and I'm guessing it's a spurious error. I've restarted that build to see if it succeeds this time.

rpavlik:
Thanks for the fix "upstream"!

I've actually got a little build fixup to do on VRPN master then I was planning on updating all the way.

I see the commit includes a change to the json-schemas - did those fall behind too, or was that an accidental revert/failure to git submodule update?

ChristophHaag:
I'm not sure how the update to json-schemas happened, but it was not intentional. Whatever revision it was updated to, it still works though.

rpavlik:
Yeah, the schemas I don't think actually get used in the build process in Core, they're just kept in there, in sync, so they can be distributed simultaneously.

rpavlik:
I'm just going to merge this one - I've got changes pending for VRPN (primarily build related stuff) but this has waited more than long enough.

Issue #441, opened 2016-06-22
Disable Client Side Prediction for HDK because of poor performances


demonixis:
Hi,

According to this [doc](https://github.com/OSVR/OSVR-Docs/blob/12e990b547586399e13d5649faf898e8826dea9f/Configuring/PredictiveTracking.md) the server side prediction is already enabled. The client side prediction is enabled too, so it can be a good idea to disable it by default for clients OR disable the server side prediction and keep the client side prediction enabled.

Today both are enabled resulting poor performances and a lag when you move your head.

russell-taylor:
I vote for disabling server-side prediction by default, because the client side actually knows how far ahead it needs to predict.

rpavlik:
Except when in extended mode, where at the moment we don't have accurate
timings, and so disabling server-side overall would be a regression.  We
already have separate configs for direct and extended mode, just need to
make the direct mode ones not use server-side prediction - I'm not sure why
this wasn't done before the client-side prediction config enabled by
default was merged in to begin with, but we can do it now in any case.

Ryan

On Thu, Jun 23, 2016 at 7:44 AM Russell Taylor <notifications@github.com>
wrote:

> I vote for disabling server-side prediction by default, because the client
> side actually knows how far ahead it needs to predict.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/OSVR/OSVR-Core/issues/441#issuecomment-228038425>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AADuyQL0fcfchWk5a1tUKgLwt1J_vxo1ks5qOn--gaJpZM4I7yPV>
> .
>
-- 
Ryan A. Pavlik, Ph.D.
CTO - OSVR Platform
Sensics, Inc.
www.sensics.com

-- 


Latest news and blog posts (subscribe here 
<http://sensics.com/subscribe-to-our-mailing-list/> to get weekly updates):

June 22: The three (and a half) configurations of eye trackers 
<http://sensics.com/the-three-and-a-half-configurations-of-eye-trackers/>

June 17: Notes from the Zero Latency free-roam VR gameplay 
<http://sensics.com/notes-from-the-zero-latency-free-roam-vr-gameplay/>

June 15: VRguy podcast: Jens Christensen, former CEO of Jaunt VR, on the 
video value chain in VR 
[...]

PR #469, opened 2016-09-09
Updated HDK Server file to disable server side prediction

demonixis:
This PR is related to the issue #441

### DIFF
--- a/apps/sample-configs/osvr_server_config.HDK12DirectMode.sample.json
+++ b/apps/sample-configs/osvr_server_config.HDK12DirectMode.sample.json
@@ -24,11 +24,5 @@
             "cameraIsForward": true
         }
     }],
-    "aliases": {
-        "/headSpace": {
-            "translate": [0.0, 0.0, 0.04141],
-            "child": "/org_osvr_filter_videoimufusion/HeadFusion/semantic/fused"
-        },
-        "/me/head": "/headSpace"
-    }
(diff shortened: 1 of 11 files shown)


rpavlik:
Thanks for this - I asked Russ for a clarification because I was under the impression that while yes, we technically are doing client side prediction in extended mode, we don't actually have the frame timing data to do it correctly there, and so in my eventual future plans of doing this one myself I was going to turn off (or perhaps just reduce to a single-eye on 1.x) client side prediction in the extended mode RM configs.

DuFF14:
I haven't compared server-side vs extended client-side , but I will put it on my list. 

Does the fusion plugin handle server-side prediction? Will removing this section impair positional tracking accuracy?
```json
"aliases": {
        "/headSpace": {
            "translate": [0.0, 0.0, 0.04141],
            "child": "/org_osvr_filter_videoimufusion/HeadFusion/semantic/fused"
        },
        "/me/head": "/headSpace"
    }
```

russell-taylor:
As I commented on Ryan's question on the issue, I think I set the extended mode prediction to be half a frame longer than the direct-mode prediction so that we're using the expected delay.  Of course, actual delay will depend on application rendering time.  The client-side timing info returns 0 in the case of no information, so kind of auto-disables itself in that case.

demonixis:
Hi, any news about this PR? Performances are critical for new users and the current configuration is not adapted.

PR #470, opened 2016-09-15
Correct VID/PID combo for 3 Space Sensor

mars979:


mars979:
@rpavlik, Actually it should have used hex literal "0x2476" instead of hex value "0x9AC" (which is 2476 in decimal)

rpavlik:
oops - yeah, vids/pids are always implied to be in hex.

rpavlik:
LGTM :+1:

PR #471, opened 2016-09-15
file logging is now disabled by default unless the OSVR_FILE_LOGGING_ENABLED environment variable is defined and not set to 0

JeroMiya:


.gitignore
```diff
@@ -1,3 +1,6 @@
+# VS Code
+.vscode/
+
```
mars979: Was this an intentional addition to the gitignore?

godbyk:
To be clear, it's not file logging per se that is the problem. It's the asynchronous logging queue which is used by all loggers.  Specifically, [this line](https://github.com/OSVR/OSVR-Core/blob/master/src/osvr/Util/LogRegistry.cpp#L209). 

The queue size refers to the number of messages the queue can hold. Right now it's set to 64k messages. We can reduce this drastically to save memory. The value must be a power of two, though (i.e., 2^x).

.gitignore
JeroMiya: Yes. Sorry, should have put it in another commit. .vscode is a temp directory from visual studio code.

Issue #474, opened 2016-09-20
Possible to load debug-runtime plugins in a non-debug-runtime server


rpavlik:
(and vice versa)

On the MSVC platform, where the C runtime (and C++ standard library, etc) are different, this is generally a no-no. I *think* it might work for plugins purely using the hourglass pluginkit APIs, but for the (bundled) plugins using the vrpnserver library, it's a quick way to end up with a strange crash. (and in any case, if you  have both, you'll likely load two copies of the same plugin with slightly different names if the default "d" postfix is added.

Have a branch that somewhat fixes things here by filtering on filenames and specifying explicitly a `.debug` postfix for plugins only (not overrideable - the reasons the postfix was desired to be overrideable in the first place do not apply to plugins) - but it may be even better to fix this (additionally) at the libfunc level, by introducing an optional additional entry point mangle `#if defined(_MSC_VER) && !defined(NDEBUG)` - it would at least keep them from running.

Easy way to reproduce this is to grab a recent core snapshot and run with an auto detect only config, and watch it go boom. Delete the plugins ending in d.dll, and it won't go boom anymore.

cc @JeroMiya since I know Android libfunc is funky too

PR #476, opened 2016-09-20
On MSVC builds, specially decorate/filter plugins built in debug

rpavlik:
Be sure we only load debug plugins iff we're a debug server.

At least partially addresses #474 - tested to fix a crasher (launching a release build in the presence of auto-loadable debug plugins) using vs14.

### DIFF
--- a/src/osvr/PluginHost/RegistrationContext.cpp
+++ b/src/osvr/PluginHost/RegistrationContext.cpp
@@ -29,19 +29,20 @@
 
 // Internal Includes
 #include <osvr/PluginHost/RegistrationContext.h>
-#include <osvr/PluginHost/SearchPath.h>
-#include <osvr/PluginHost/PathConfig.h>
+
 #include "PluginSpecificRegistrationContextImpl.h"
-#include <osvr/Util/Verbosity.h>
+#include <osvr/PluginHost/PathConfig.h>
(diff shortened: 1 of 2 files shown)


src/osvr/PluginHost/RegistrationContext.cpp
```diff
@@ -50,6 +51,9 @@
 namespace osvr {
 namespace pluginhost {
     static const auto PLUGIN_HOST_LOGGER_NAME = "PluginHost";
+#ifdef _MSC_VER
+    static const auto PLUGIN_HOST_DEBUG_SUFFIX = ".debug";
```
godbyk: Why not set `PLUGIN_HOST_DEBUG_SUFFIX` to `""` on non-MSVC platforms to remove an `#ifdef` block later?

godbyk:
See inline comment, but otherwise looks good to me. :+1:

src/osvr/PluginHost/RegistrationContext.cpp
rpavlik: because we don't need to filter plugins on other platforms, only on MSVC because it has two different C runtimes?

rpavlik:
We do exclude them, AFAIK - this bug is seen when using a core snapshot or SDK build that includes a debug build.

PR #478, opened 2016-09-20
Removed unused and undefined VRPNMultiserverData::assignName() function.

godbyk:
cc: @rpavlik in case he had plans for this function.


rpavlik:
If I did, I don't remember it. assuming travis builds it OK, can merge this with the next batch of PRs.

PR #481, opened 2016-10-03
Replaced regex call with std::string methods and isdigit() to avoid c…

russell-taylor:
…rash on some compiler versions in Linux.

### DIFF
--- a/src/osvr/Common/PathElementTypes.cpp
+++ b/src/osvr/Common/PathElementTypes.cpp
@@ -30,7 +30,6 @@
 // - none
 
 // Standard includes
-#include <regex>
 #include <sstream>
 
 namespace osvr {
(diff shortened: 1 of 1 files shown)


src/osvr/Common/PathElementTypes.cpp
```diff
@@ -89,9 +88,18 @@ namespace common {
         /// port if one isn't already specified.
         static inline std::string
         attachPortToServerIfNoneSpecified(std::string const &server, int port) {
-            auto endsWithPort = std::regex{".*:[[:digit:]]+$"};
```
godbyk: Does replacing `[[:digit:]]` with `[0-9]` work, by chance?

The underlying problem is that gcc 4.8 doesn't have full regex support.

russell-taylor: Tried that and it didn't work.  Thanks for the idea.

russell-taylor:
It was the only use, I thought I pulled out the #include of it...

russell-taylor:
Must have done that in an earlier, abandoned attempt.

russell-taylor:
Yeah, it got removed in this request as well.  Thanks.

PR #484, opened 2016-10-05
Various logging updates

godbyk:
- Updated spdlog to incorporate some improvements to the async queue behavior. This will hopefully fix the 'stall on `FreeLibrary()` issue.
- Fixed bug where `logger->info("something")` would print the pointer address instead of the C string.
- Reduced async log queue size to 64 messages. This should drastically improve the memory usage of `osvr_server` at the potential expense of latency in logging messages. If the latency becomes a problem, there are further solutions we can employ.
- Added `drop()` and `dropAll()` functions to unregister a logger with the LogRegistry. If no one else is using the logger, it will be destroyed, flushing its queue and closing its log file in the process.

### DIFF
--- a/inc/osvr/Util/Log.h
+++ b/inc/osvr/Util/Log.h
@@ -75,13 +75,13 @@ namespace util {
         /// The logger will survive until the last copy of it is destroyed
         /// (e.g., goes out of scope). This function is useful if you want to
         /// destroy a logger before the program terminates.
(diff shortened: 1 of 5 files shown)


rpavlik:
This appears to fix the "printing string literals as pointers" and also fix, or at least hide, a logger construction race condition (which may be something we want to add a test for).

PR #485, opened 2016-10-06
Moved type definitions out of std:: namespace so the code will compil…

russell-taylor:
…e on VS 2013.


godbyk:
The `std::` namespace is fine, you just need to `#include <cstdint>`.

rpavlik:
Sorry I missed that include apparently.

PR #494, opened 2016-10-21
Fix for debug mode keys on linux platforms

toastedcrumpets:
In short, cv::waitKey() returns scancodes (not ascii chars) on non-windows platforms, allowing use of non-ascii jeys (i.e. caps lock, arrow keys etc); However, this means the current code is broken on linux as it cannot correctly identify key presses in two key locations, as most users have numlock on, resulting in a high bit being set in the scankey.

This pull request masks the scancode so any modifier bits are filtered out, making it compatible with char/ascii types.

This fixes the debug window for me on Linux.


rpavlik:
ah, great, thank you!

PR #495, opened 2016-10-24
Update to Boost 1.62.0.

godbyk:
There were no substantial updates to Boost IPC.

Is this version check still worth doing? Or can we just blacklist the known-bad version(s)?


rpavlik:
It's just a question of policy posture: whether we trust boost.ipc to not break its abi.

PR #498, opened 2016-10-28
uninstall numpy installed by pip before installing opencv

mars979:
Uninstall numpy that was installed by pip before installing opencv, otherwise osx build fails. 
Found this issue that was filed recently - https://github.com/travis-ci/travis-ci/issues/6688 where numpy comes pre-installed by pip, so it conflicts when you do brew install opencv
It caused all the recent Travis builds on OSX to fail, so this should fix the problem


rpavlik:
Looks good, thank you for figuring this out!

PR #505, opened 2016-12-07
Add missing include to IPCBuffer.cpp

godbyk:
Need to include `Logger.h` otherwise it's only a forward-declaration and fails to compile.


rpavlik:
How did this build before, then?

godbyk:
@rpavlik No idea. Works fine for me locally. But that's one of the errors I'm getting on Travis CI.

rpavlik:
LGTM

PR #511, opened 2016-12-21
Fixes bug in error reporting during loading of plugins

godbyk:
Bug reported via support ticket #1235.


mars979:
Ah! Nice catch, looks like a left over copy paste. LGTM 👍

Issue #31, opened 2015-02-27
Add helpful check to CMake to ensure submodules are available


rpavlik:
Right now it just complains without explaining what to do (at least with the json-schemas submodule) to fix it.

Issue #30, opened 2015-02-25
Byte order handling


rpavlik:
Leaving links as notes.

https://wiki.debian.org/ArmEabiFixes#Floating_point_format_has_changed

PR #518, opened 2017-01-08
Update for Boost 1.63

stanlarroque:
Homebrew uses Boost 1.63, therefore OSVR-Core doesn't compile on MacOS.

### DIFF
--- a/cmake-local/BoostTargets.cmake
+++ b/cmake-local/BoostTargets.cmake
@@ -23,8 +23,8 @@ else()
 endif()
 
 find_package(Boost ${OSVR_MIN_BOOST} COMPONENTS ${required_boost_components} REQUIRED)
-if(Boost_VERSION GREATER 106200)
-    # Current max code-reviewed version: Boost 1.62
+if(Boost_VERSION GREATER 106300)
+    # Current max code-reviewed version: Boost 1.63
     # When this is updated - source code must also be updated!
     message(SEND_ERROR "Using an unreviewed Boost version - inspect the Boost Interprocess release notes/changelog/diffs to see if any ABI breaks took place as they may affect client/server interoperability.")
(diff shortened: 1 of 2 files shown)


godbyk:
The changes in Boost.IPC 1.63.0 are:

 * [Trac #12499 ("Memory allocation fails")](https://svn.boost.org/trac/boost/ticket/12499)
 * [GitHub Pull #30 ("Provide extension point so library user can provide default temp folder")](https://github.com/boostorg/interprocess/pull/30)
 * [GitHub Pull #31 ("Add xsi_key(key_t) constructor")](https://github.com/boostorg/interprocess/pull/31)

The second one (PR #30) is good to keep in mind if we need to make a sandboxed application on macOS as the `/tmp` directory isn't writable to sandboxed apps there.

rpavlik:
@godbyk do any of the changes affect the IPC ABI?

godbyk:
I haven't tested it directly, but here's my (visual) review of the ABI changes:

 * [Trac #12499](https://svn.boost.org/trac/boost/ticket/12499) I'm not sure about. It makes a bunch of changes. All the changes are in the Boost.Move headers but I'm not sure which headers are actually used by IPC.
 * https://github.com/boostorg/interprocess/pull/30 looks to be ABI stable. 
 * https://github.com/boostorg/interprocess/pull/31 looks to be ABI stable.

Somewhat unrelated, but I found a nice [ABI report](https://abi-laboratory.pro/tracker/objects_report/boost/1_62_0/1_63_0/report.html) comparing the 1.63.0 libraries to 1.62.0. Unfortunately, it doesn't show me the header-only libraries. Looks like Boost.Regex is the only one that has backward-compatibility issues.

rpavlik:
The only ABI I care about is the "wire" ABI - whether 1.62 clients can talk to 1.63 servers and vice-versa. I am getting sick of these PRs - I think I'm about worn down to the point of accepting just a warning instead of an error on new Boost versions we didn't test :D

PR #522, opened 2017-02-01
Tracking refinement

rpavlik:
Some cleanup and improvements based on the initial feedback from Reddit and others. Reduces the number of config directives in the sample config files, improves messages (removes the EXPERIMENTAL warning from startup), and improves the RenderManager config files.

### DIFF
--- a/apps/sample-configs/renderManager.direct.portrait.json
+++ b/apps/sample-configs/renderManager.direct.portrait.json
@@ -6,10 +6,10 @@
         "directModeEnabled": true,
         "directDisplayIndex": 0,
         "directHighPriorityEnabled": true,
-        "numBuffers": 1,
-        "verticalSyncEnabled": false,
+        "numBuffers": 2,
+        "verticalSyncEnabled": true,
         "verticalSyncBlockRenderingEnabled": true,
-        "renderOverfillFactor": 1.5,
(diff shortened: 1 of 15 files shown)


apps/sample-configs/renderManager.direct.portrait.json
```diff
@@ -6,10 +6,10 @@
         "directModeEnabled": true,
         "directDisplayIndex": 0,
         "directHighPriorityEnabled": true,
-        "numBuffers": 1,
-        "verticalSyncEnabled": false,
+        "numBuffers": 2,
```
JeroMiya: just curious - is front-buffer rendering even expected to work in the current RenderManager? I haven't been able to get it to work without tearing so far.

rpavlik: I don't know. Other vendors don't do it.

russell-taylor: Getting front-buffer tearing to work would be a remarkable feat.  It requires timing things so that you do all of your rendering and buffer swapping during vertical blanking.  It might some day conceivably work when we're using copy buffers.  Note that double buffering in this context does not provide additional latency -- we're still rendering right before swapping the buffers.

PR #529, opened 2017-03-09
Added osvr_server command-line options

godbyk:
This PR adds command-line options to the server for adjusting the log verbosity and also allows the server to run without specifying a configuration file (in which case it will assume an empty configuration `{ }`).

```
$ ./osvr_server --help
Command Line Options:
  --help                display this help message
  -v [ --verbose ]      enable verbose logging
  -d [ --debug ]        enable debug logging
```

Verbose logging sets the log level to `debug` and debug logging sets the log level to `trace`.

This subsumes PR #293 by @Outurnate.

### DIFF
--- a/apps/osvr_server.cpp
+++ b/apps/osvr_server.cpp
@@ -30,12 +30,18 @@
 #include <osvr/Util/LogNames.h>
 
 // Library/third-party includes
-// - none
+#include <boost/program_options.hpp>
+#include <boost/filesystem.hpp>
+#include <boost/filesystem/fstream.hpp>
+#include <boost/optional.hpp>
 
 // Standard includes
 #include <exception>
 #include <fstream>
 #include <iostream>
+
+namespace opt = boost::program_options;
+namespace fs = boost::filesystem;
 
 static osvr::server::ServerPtr server;
 using ::osvr::util::log::OSVR_SERVER_LOG;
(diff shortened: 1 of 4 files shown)


apps/osvr_server.cpp
```diff
@@ -50,16 +57,84 @@ void handleShutdown() {
...
-    std::string configName(osvr::server::getDefaultConfigFilename());
-    if (argc > 1) {
-        configName = argv[1];
+    std::string configName; // server configuration filename
+
+    opt::options_description optionsAll("All Options");
+    opt::options_description optionsVisible("Command Line Options");
+    opt::positional_options_description optionsPositional;
+
+    optionsPositional.add("config", -1);
```
JeroMiya: Does this make the config option the default option when only one argument is passed (like when you drag a config file over the server in windows explorer)? Just making sure we don't change the current behavior in that case.

apps/osvr_server.cpp
```diff
@@ -50,16 +57,84 @@ void handleShutdown() {
...
+        osvr::util::log::LogRegistry::instance().setLevel(osvr::util::log::LogLevel::trace);
+        osvr::util::log::LogRegistry::instance().setConsoleLevel(osvr::util::log::LogLevel::trace);
+        log->trace("Debug logging enabled.");
+    } else if (values.count("verbose")) {
+        osvr::util::log::LogRegistry::instance().setLevel(osvr::util::log::LogLevel::debug);
+        osvr::util::log::LogRegistry::instance().setConsoleLevel(osvr::util::log::LogLevel::debug);
+        log->debug("Verbose logging enabled.");
+    }
+
+    configName = values["config"].as<std::string>();
```
JeroMiya: what does this line do if no command line arguments are passed?

apps/osvr_server.cpp
```diff
@@ -50,16 +57,84 @@ void handleShutdown() {
...
+        configPath = boost::none;
+    }
+
+    if (configPath) {
+        server = osvr::server::configureServerFromFile(configName);
     } else {
-        log->info()
-            << "Using default config file - pass a filename on the command "
-               "line to use a different one.";
+        server = osvr::server::configureServerFromString("{ }");
```
JeroMiya: We need to also check for a file called `osvr::server::getDefaultConfigFilename()` in the current working directory, so that we don't break server auto-start and double-click behavior (i.e. starting with no arguments should load the default config file name from the CWD). Only if the config is not specified, and the default-named config file is not found in the CWD, should we be using a blank config as is done here.

apps/osvr_server.cpp
godbyk: Yes. So if you run `osvr_server blah.json` then it will use `blah.json` as the configuration file. The same thing happens if you drag and drop `blah.json` onto the `osvr_server.exe` program in Windows.

apps/osvr_server.cpp
godbyk: @JeroMiya If no configuration file is specified, it will check to see if the file returned by `osvr::server::getDefaultConfigFilename()` exists and is readable. If it is, then we'll use it. Otherwise, we'll start the server with an empty `{ }` configuration.

apps/osvr_server.cpp
godbyk: If no configuration file is specified on the command line, we assume the user specified the filename returned by `osvr::server::getDefaultConfigFilename()`. If that file exists, we'll use it. If it doesn't, then we'll start the server with an empty `{ }` configuration.

godbyk:
If no configuration file is specified, we [default to using `osvr::server::getDefaultConfigFilename()`](https://github.com/godbyk/OSVR-Core/blob/logging-improvements/apps/osvr_server.cpp#L69). If that fails (e.g., that file doesn't exist), then we'll use an empty `{ }` configuration.

apps/osvr_server.cpp
```diff
@@ -50,16 +57,84 @@ void handleShutdown() {
...
+                return -1;
+            }
+        }
+    } catch (fs::filesystem_error &e) {
+        log->error() << "Could not open config file at '" << configName << "'.";
+        log->error() << "Reason " << e.what() << ".";
+        configPath = boost::none;
+    }
+
+    if (configPath) {
```
mars979: I'm assuming this check fails if you provide an empty string to `boost::optional<fs::path> configPath(configName)` on line 111 ?

inc/osvr/Server/ConfigureServerFromFile.h
```diff
@@ -48,24 +49,15 @@ namespace server {
...
-        }
-
         osvr::server::ConfigureServer srvConfig;
         log->info() << "Constructing server as configured...";
         try {
-            srvConfig.loadConfig(config);
+            srvConfig.loadConfig(json);
             ret = srvConfig.constructServer();
         } catch (std::exception &e) {
             log->error()
```
mars979: Should we log the std::exception error here ?

godbyk: @mars979 We are logging the exception here. (Or are you suggesting we _shouldn't_ log it?)

apps/osvr_server.cpp
godbyk: @mars979 I think so, since I would expect `fs::exists()` to return false on an empty string.

mars979: Ok, sounds good.

inc/osvr/Server/ConfigureServerFromFile.h
mars979: @godbyk , I meant to ask if `log->error()` automatically logs the `std::exception &e` or do you need to explicitly add `log->error(e.what())` (or something similar)

godbyk: @mars979 Ah, gotcha. We're logging it a couple lines down. (clang-format's line-wrapping hinders the readability of the line.)

mars979: Ah, it was just wrapped to next line. My bad.

PR #541, opened 2017-05-11
Android config path

JeroMiya:


src/osvr/Server/ConfigFilePaths.cpp
```diff
@@ -77,20 +77,26 @@ namespace osvr {
             configDir /= fs::path("OSVR") / configSubpath;
 #endif
 
-            return configDir.string();
+#if defined(OSVR_ANDROID)
+            configDir = fs::path("/sdcard/osvr");
+#endif
+            auto ret = configDir.string();
```
rpavlik: Any particular reason you split this up? (Does `.string()` return a reference or something weird?)  It's not a blocker, just curious.

JeroMiya: No particular reason. I was diagnosing a stdlib error in this code and tried splitting it up to see if it was happening in string() and left the change in. OK to leave in, or should I revert it?

PR #544, opened 2017-05-19
Small android-specific fix for plugins to allow them to work on API level 25

JeroMiya:


cmake-local/osvrAddPlugin.cmake
```diff
@@ -22,7 +22,15 @@ function(osvr_add_plugin)
     if(NOT OSVR_ADD_PLUGIN_SOURCES)
         set(OSVR_ADD_PLUGIN_SOURCES ${OSVR_ADD_PLUGIN_UNPARSED_ARGUMENTS})
     endif()
-    add_library(${OSVR_ADD_PLUGIN_NAME} MODULE ${OSVR_ADD_PLUGIN_SOURCES})
+    
+    if(ANDROID)
+        # Android loads plugins as if they were shared libraries, and
+        # additionally, requires SONAME to be set.
+        add_library(${OSVR_ADD_PLUGIN_NAME} SHARED ${OSVR_ADD_PLUGIN_SOURCES})
```
rpavlik: Please make a `TODO` note this is a hack/workaround for a defect in the toolchain configuration file: I don't think this should be needed, but it's the easiest way for you to get the behavior you want right now.

JeroMiya: I added a TODO note.

PR #545, opened 2017-05-25
fix gcc compiler error about for loop initial declaration

mars979:


godbyk:
If you instead add the following to the `CMakeLists.txt` file, does it compile as is?

```cmake
set(CMAKE_C_STANDARD 11) # 99 or 03 or whichever standard you're aiming for
set(CMAKE_C_STANDARD_REQUIRED ON)
```

mars979:
@godbyk ,
The client examples are compiled all together in CMakeLists, so it would be applied to all of them at once. Ryan brought up a point that MSVC C compiler doesn't support 99 for example, so that's why I'm just fixing this one example.

godbyk:
@mars979 MSVC C compiler doesn't support _all_ of C99 (or all of _any_ C standard, as far as I'm aware), but it does support initializations in for loops.

I'm not concerned about this single example, but about solving the problem going forward.

PR #546, opened 2017-05-26
remove boost version check

ChristophHaag:
There once was a boost release that broke some ABI.
This has not happened again in a looong time, so remove the inconvenience of having to update this check after every boost release.

In my opinion it would be far better to document the faulty boost version in a "known issues" section in the Readme instead of having this rather pointless check in the code and failing the build.

If this is not acceptable, maybe the compile error could at least be made into a warning?


rpavlik:
Thanks - they hadn't broken it in a long time, several years now, and you're right - we're effectively breaking things more often than boost because of this build setup.

PR #550, opened 2017-06-14
Fix for plugins that are built as debug or as manualload debug

mars979:
When plugin is built as debug or as manulload debug you would run into this scenarios:
1. Plugins named "pluginName.debug.dll" would get extra ".dll.debug" added because PluginHost thinks that pluginBaseName is just "pluginName"
2. Loading Debug Manualload plugins by name such as "pluginName.debug.manualload.dll" is treated as "pluginName.manualload.debug".

This PR fixes the issue by avoiding appending extra extensions for debug plugins and use the correct order for manualload debug plugins.


src/osvr/PluginHost/RegistrationContext.cpp
```diff
@@ -51,9 +51,6 @@
 namespace osvr {
 namespace pluginhost {
     static const auto PLUGIN_HOST_LOGGER_NAME = "PluginHost";
-#ifdef _MSC_VER
```
godbyk: If the `.debug` suffix is no longer MSVC-only, then you should also modify `cmake-local/osvrAddPlugin.cmake` accordingly (lines 45–48).

src/osvr/PluginHost/RegistrationContext.cpp
```diff
@@ -175,7 +167,7 @@ namespace pluginhost {
             // Visual C++ debug runtime: we append to the plugin name. Must only
             // load debug plugins iff we're a debug server
             const auto isDebugRuntimePlugin =
-                boost::iends_with(pluginBaseName, PLUGIN_HOST_DEBUG_SUFFIX);
+                boost::iends_with(pluginBaseName, OSVR_PLUGIN_DEBUG_SUFFIX);
```
godbyk: I don't know the answer to this, but if we have two suffixes (`.debug` and `.manualload`), which order must they appear in if they're both used? Is that order imposed properly by the build tools? Do we want to support allowing the suffixes to appear in any order?

src/osvr/PluginHost/RegistrationContext.cpp
godbyk: See `src/osvr/PluginHost/RegistrationContext.cpp`, circa line 168 where we check to see if the `.manualload` suffix appears at the end of the name.

src/osvr/PluginHost/SearchPath.cpp
```diff
@@ -172,11 +172,20 @@ namespace pluginhost {
                 }
                 const auto pluginBaseName =
                     pluginCandidate.filename().stem().generic_string();
+
+#if defined(_MSC_VER) && !defined(NDEBUG)
+                // Visual C++ debug runtime: we append to the plugin name.
+                const std::string decoratedPluginName =
+                    pluginName + OSVR_PLUGIN_DEBUG_SUFFIX;
+#else
+                const std::string &decoratedPluginName = pluginName;
```
godbyk: Why is `decoratedPluginName` a reference here but not on line 178 above? (Actually, I know the answer. But it's probably better to store a copy in both cases for consistency's sake.)

src/osvr/PluginHost/SearchPath.cpp
```diff
@@ -172,11 +172,20 @@ namespace pluginhost {
...
+#endif
+
                 /// If the name is right or has the manual load suffix, this is
                 /// a good one.
                 if ((pluginBaseName == pluginName) ||
```
godbyk: Shouldn't `pluginName` here also be `decoratedPluginName`?

rpavlik:
@mars979 is this still an issue/open? Any response to the review comments?

src/osvr/PluginHost/RegistrationContext.cpp
mars979: Based on the latest OSVR Core snapshot, I'm seeing that plugins that are debug and manualload are named as follows: `com_osvr_pluginName.debug.manualload.dll` 

From  osvrAddPlugin.cmake script I confirmed that 
1. Plugin `SUFFIX` is set to `manualload.dll` (if it is a manualload plugin) which gets appended to the end of the target name (according to the <https://cmake.org/cmake/help/v3.6/prop_tgt/SUFFIX.html>)
2. Plugin `DEBUG_POSTIFX` is set to `.debug` which is appended to the target file name (https://cmake.org/cmake/help/v3.6/prop_tgt/CONFIG_POSTFIX.html)
Hence the order would always be as follows `com_osvr_pluginName.debug.manualload.dll` for debug, manualload plugins.

src/osvr/PluginHost/SearchPath.cpp
mars979: That's probably because it's a copypasta from its original location. I moved this piece of code from RegistrationContext.cpp. Good catch! I'll change that

PR #556, opened 2017-07-22
Just in timewarp

russell-taylor:
Adds entries into the RenderManager configuration to control a new feature.  Also adds an inverse function that was used by that new code.

Would love to get this in before the VR Gems book chapter describing it goes to the publisher next week...


russell-taylor:
Going to pull the display information into another class.  Will re-issue a new request for just the inverse function.

PR #559, opened 2017-08-02
Just in timewarp

russell-taylor:
This has been modified to only include the specification of whether we are using just-in-timewarp, the scan orientation has been moved to OSVR-RenderManager.

This must be merged before we can merge the just-in-timewarp branch from RenderManager.


russell-taylor:
Looks like this is clear to merge without review, which surprises me given that RenderManager's requires a review.  I'll go ahead and merge it tomorrow unless someone asks me to hold off.

PR #561, opened 2017-08-07
Build fixes

rpavlik:
These are just some basic cleanups for VS2017 that I apparently had laying around. I have split out the commit that adjusts the boost helper to accept boost built for vs2015 when building with vs2017, so that's not here, since that's presumably more controversial.  (It also includes the latest updates from https://github.com/rpavlik/cmake-modules some of which are related.)


cmake/CppcheckTargets.cmake
```diff
@@ -30,16 +34,24 @@ if(NOT CPPCHECK_FOUND)
...
+	set_target_properties(all_cppcheck PROPERTIES EXCLUDE_FROM_ALL TRUE)
+elseif(CPPCHECK_VERSION VERSION_LESS 1.53.0)
+	add_custom_target(all_cppcheck
+		COMMENT "Need at least cppcheck 1.53, found ${CPPCHECK_VERSION}")
+	set_target_properties(all_cppcheck PROPERTIES EXCLUDE_FROM_ALL TRUE)
+	set(CPPCHECK_FOUND)
+endif()
+
+if(NOT TARGET all_cppcheck)
+  add_custom_target(all_cppcheck)
```
godbyk: Inconsistent indentation, but should probably be corrected upstream.

cmake/FindGLEW.cmake
```diff
@@ -0,0 +1,132 @@
...
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#=============================================================================
+
+find_path(GLEW_INCLUDE_DIR GL/glew.h)
+if(WIN32)
+  # TODO how to make this exclude arm?
```
godbyk: You could be able to use `if(CMAKE_SYSTEM_PROCESSOR MATCHES "arm")` to detect ARM systems.

cmake/InstallDebugSymbols.cmake
```diff
@@ -0,0 +1,85 @@
...
+                endif()
+            endif()
+            install(FILES $<${HAS_SYMBOLS_CONDITION}:$<TARGET_PDB_FILE:${_target}>>
+                DESTINATION ${DEST}
+                ${EXTRA_INSTALL_ARGS})
+        endforeach()
+    endfunction()
+else()
+    function(install_debug_symbols)
+        # do nothing if too old of CMake or not MSVC.
```
godbyk: Perhaps emit a dev warning if CMake is too old so it doesn't silently fail?

inc/osvr/Util/Logger.h
```diff
@@ -203,7 +203,7 @@ namespace util {
             OSVR_UTIL_EXPORT void flush();
 
             /// Get the logger name
-            OSVR_UTIL_EXPORT std::string const &getName() const {
```
godbyk: Why should we not export `Logger::getName()`?

rpavlik: it's inline

godbyk: Right. I think we addressed all the Logger export stuff in a separate PR already.

PR #566, opened 2017-09-28
Added moverio sample configs

JeroMiya:
osvr_server_config.moverio.json is the all-in-one configuration (display config and render manager config) for the moverio using its native FOV. Good for AR applications.

osvr_server_config.moverio_highfov.json is the all-in-one configuration that includes a larger FOV (but retaining the correct aspect ratio) and is more comfortable for VR style applications.

### DIFF
--- a/apps/sample-configs/osvr_server_config.moverio.json
+++ b/apps/sample-configs/osvr_server_config.moverio.json
@@ -1,5 +1,5 @@
 {
-  "description": "This is a native FOV configuration for the Moverio BT-300 series AR devices.",
+  "description": "This is a native FOV configuration for the Moverio BT-300 series AR devices. This is a speciallized configuration that is more accurate for scale, but may result in visibility issues due to low FOV. The highfov configuration is recommended for most purposes.",
   "#": "The display and renderManagerConfig sections below are embedded as a workaround to missing relative path functionality on Android.",
   "display": {
     "meta": {
@@ -9,8 +9,7 @@
       "device": {
         "vendor": "Epson",
         "model": "Moverio BT-300 series",
-        "num_displays": 1,
-        "Note": "This configuration represents the native FOV of the Moverio BT-300 series. This is a speciallized configuration that is more accurate for scale, but may result in visibility issues due to low FOV. The highfov configuration is recommended for most purposes."
+        "num_displays": 1
       },
       "field_of_view": {
         "monocular_horizontal": 20.121542144531,
--- a/apps/sample-configs/osvr_server_config.moverio.json
+++ b/apps/sample-configs/osvr_server_config.moverio.json
@@ -1,98 +1,82 @@
 {
-  "description": "This configuration supports video (so-called 'positional') and IMU fusion tracking, in addition to orientation-only tracking, with the OSVR HDK 2, for units that were manufactured as HDK2 (NOT UPGRADED FROM HDK 1.x!). It is configured for RenderManager applications in direct mode on HDK2 optics.",
+  "description": "This is a native FOV configuration for the Moverio BT-300 series AR devices.",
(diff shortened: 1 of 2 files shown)


apps/sample-configs/osvr_server_config.moverio.json
```diff
@@ -0,0 +1,98 @@
...
+
+        "timeWarp": {
+            "enabled": false,
+            "asynchronous": false,
+            "maxMsBeforeVSync": 4
+        }
+    }
+  },
+  "plugins": [
+    "org_osvr_android_moverio", "com_osvr_android_jniImaging"
```
mars979: Are these manual load plugins ?

JeroMiya: All plugins in a monolithic deployment (e.g. Android) are manual load.

apps/sample-configs/osvr_server_config.moverio.json
```diff
@@ -0,0 +1,98 @@
...
+  "display": {
+  "meta": {
+    "schemaVersion": 1
+  },
+  "hmd": {
+    "device": {
+      "vendor": "Epson",
+      "model": "Moverio BT-300 series",
+      "num_displays": 1,
+      "Version": "1.0",
```
rpavlik: This is actually hardware version, IIRC - I think you can actually leave it out if it's not meaningful.

apps/sample-configs/osvr_server_config.moverio.json
```diff
@@ -0,0 +1,98 @@
+{
```
rpavlik: General comment: I'd suggest editing this in a JSON-aware editor that can beautify (basically, keep indents nice or make them nice), or at least run it through something like jsbeautify.

apps/sample-configs/osvr_server_config.moverio_highfov.json
```diff
@@ -0,0 +1,98 @@
...
+  "meta": {
+    "schemaVersion": 1
+  },
+  "hmd": {
+    "device": {
+      "vendor": "Epson",
+      "model": "Moverio BT-300 series",
+      "num_displays": 1,
+      "Version": "1.0",
+      "Note": "This is not the native FOV of the device but a higher FOV for 'magic window' style display."
```
rpavlik: This is not how I understand the term magic window: as I understand it, is more like ARCore: viewing surface in the world that's tracked and has typically has a camera (video see-thru), not head-coupled.

apps/sample-configs/osvr_server_config.moverio_highfov.json
```diff
@@ -0,0 +1,98 @@
...
+    },
+    "field_of_view": {
+      "monocular_horizontal": 90,
+      "monocular_vertical": 50.98299243,
+      "overlap_percent": 100,
+      "pitch_tilt": 0
```
rpavlik: `"pitch_tilt": 0` is default so can be omitted.

apps/sample-configs/osvr_server_config.moverio_highfov.json
```diff
@@ -0,0 +1,98 @@
...
+    },
+    "resolutions": [
+      {
+        "width": 1280,
+        "height": 720,
+        "video_inputs": 1,
+        "display_mode": "horz_side_by_side",
+        "swap_eyes": 0
```
rpavlik: this line is default and can be omitted. Same comments from previous file apply to this as well.

apps/sample-configs/osvr_server_config.moverio_highfov.json
JeroMiya: Imagine ARCore style magic window, but instead of holding it with your hand, you have your phone strapped to your head and held about half a foot in front of your face. That's what the Moverio looks like. If you have a better term for it, I'll use it.

PR #574, opened 2017-10-31
Remove export declarations from deleted/inline functions

### LINKED ISSUE #570: Fail to compile on mingw32 gcc 7.2
Master branch doesn't compile on msys2/mingw32, with gcc 7.2:

[ 12%] Building CXX object src/osvr/Common/CMakeFiles/osvrCommon.dir/AliasProcessor.cpp.obj
In file included from C:/build/OSVR-Core/build/src/osvr/Util/Verbosity.h:43:0,
                 from C:\build\OSVR-Core\src\osvr\Common\AliasProcessor.cpp:32:
C:/build/OSVR-Core/inc/osvr/Util/Logger.h:206:49: error: function 'const string& osvr::util::log::Logger::getName() const' definition is marked dllimport
             OSVR_UTIL_EXPORT std::string const &getName() const {
                                                 ^~~~~~~
mingw32-make[2]: *** [src\osvr\Common\CMakeFiles\osvrCommon.dir\build.make:88: src/osvr/Common/CMakeFiles/osvrCommon.dir/AliasProcessor.cpp.obj] Error 1
mingw32-make[1]: *** [CMakeFiles\Makefile2:1729: src/osvr/Common/CMakeFiles/osvrCommon.dir/all] Error 2
mingw32-make: *** [Makefile:162: all] Error 2

rpavlik:
This was sitting in my VS2017 branch, but I imagine it would also fix #570 . Pretty simple: don't import/export from DLL functions that are deleted or inline.

### DIFF
--- a/inc/osvr/Util/LogRegistry.h
+++ b/inc/osvr/Util/LogRegistry.h
@@ -58,10 +58,10 @@ namespace util {
 
         class LogRegistry {
           public:
-            OSVR_UTIL_EXPORT LogRegistry(LogRegistry const &) = delete; // copy construct
-            OSVR_UTIL_EXPORT LogRegistry(LogRegistry &&) = delete;      // move construct
-            OSVR_UTIL_EXPORT LogRegistry &operator=(LogRegistry const &) = delete; // copy assign
-            OSVR_UTIL_EXPORT LogRegistry &operator=(LogRegistry &&) = delete;      // move assign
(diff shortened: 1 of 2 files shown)


rpavlik:
Well, I guess GCC 7 does, and based on where I found this commit laying around, apparently VC2017 does?

rpavlik:
Added a few more identical fixes when I actually tried an msys2 build myself, will wait for the travis build.

PR #577, opened 2017-11-17
Reset yaw

DuFF14:
Adds osvrResetYaw API call


src/osvr/CMakeLists.txt
```diff
@@ -33,3 +33,5 @@ endif()
 if(BUILD_JOINTCLIENTKIT)
     add_subdirectory(JointClientKit)
 endif()
+
+add_subdirectory(ResetYaw)
```
godbyk: Should this be wrapped in `if(BUILD_RESETYAW) ... endif()`?

PR #586, opened 2018-07-05
Cleanup

rpavlik:
Small cleanup of things seen when building on Linux with Clang 6.0. (Plus one or two things that I just noticed could be improved as drive-by fixes)

I should try this on Windows before merging it, in theory.

### DIFF
--- a/inc/osvr/Common/CommonComponent.h
+++ b/inc/osvr/Common/CommonComponent.h
@@ -87,6 +87,12 @@ namespace common {
         /// side of a DLL line.
         static OSVR_COMMON_EXPORT shared_ptr<CommonComponent> create();
 
+        /// @brief Explicit virtual destructor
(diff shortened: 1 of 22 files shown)


godbyk:
`= default` could be declared in the headers unless you want to hide the implementation. 

Otherwise, it looks good to me on first read through. I haven't actually tested it.

rpavlik:
Have to hide the implementation to ensure the destruction happens within a dll, not in its consumer - for the same reason that the constructor and factory func are hidden. Mainly a very sneaky windows bug.

rpavlik:
OK, this has been rebased, so travis should work now. Still haven't tried Windows (and AppVeyor isn't set up right)

rpavlik:
OK, doesn't appear to break Windows, so merging.

PR #493, opened 2016-10-20
uvc-camera for blobs-undo-bad

ChristophHaag:
Commits from https://github.com/OSVR/OSVR-Core/pull/492

Not ready for merging, but this may be used as a reference.

PR #587, opened 2018-07-06
Libuvc in new tracker plugin

rpavlik:
Thanks to @ChristophHaag - this is just the same work as #493 rebased and squashed, with a little bit of build system improvement.

It does launch and run, though for some reason it's dropping frames once it gets an eye on the device, even on a very beefy computer - haven't looked into why yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,9 +103,20 @@ if(BUILD_SERVER_PLUGINS AND BUILD_WITH_OPENCV)
             endif()
         endif()
     elseif(NOT ANDROID)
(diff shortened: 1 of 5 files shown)


ChristophHaag:
Thanks for fixing this up. :)

I remember I wanted to look into something with a timestamp, whether it was taken/used at the correct point in time and whether it may be the cause of possible problems.

When I last tried it, the performance wasn't bad though.

rpavlik:
OK, this has been updated to be based on the latest, travis-fixed, master branch. Good to merge if it doesn't break anything, I think.

PR #590, opened 2018-07-09
Add udev rules

rpavlik:
Not sure why I didn't commit these long ago - most of them have been sitting around in a Gist since 2015.

No build system touches them yet but this at least puts them in the source tree.

**Question** - is `MODE="0660" GROUP="plugdev"` the best thing to do to provide USB access? It's different from what the Vive linux support does - https://github.com/ValveSoftware/SteamVR-for-Linux#usb-device-requirements

### DIFF
--- a/udev-rules.d/99-osvr-hdk-camera.rules
+++ b/udev-rules.d/99-osvr-hdk-camera.rules
@@ -1,0 +1,6 @@
+# udev file for OSVR HDK positional tracking camera
+# Copyright 2018 Collabora
+# SPDX-License-Identifier: Apache-2.0
+
+# OSVR HDK Camera (via libusb/libuvc)
+SUBSYSTEM=="usb", ATTRS{idVendor}=="0bda", ATTRS{idProduct}=="57e8", MODE="0660", GROUP="plugdev"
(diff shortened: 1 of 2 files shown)


godbyk:
Can you think of any benefit to restricting access to the device to a subset of users on a system?

The [Debian wiki](https://wiki.debian.org/SystemGroups) says that the `plugdev` group "Allows members to `mount` (only with the options `nodev` and `nosuid`, for security reasons) and `umount` removable devices through `pmount`."

The [Ubuntu wiki](https://wiki.ubuntu.com/Security/Privileges) says the group provides "access external storage devices" and that "users in the `plugdev` group can send commands to HAL (this is probably deprecated)."

Other distributions seem to lack this group, so it's probably easiest to just omit the group entirely from the udev rule, allowing all users to access the devices.

rpavlik:
Well, it makes it so if you have somebody that only logs in remotely, you
can leave them out of that group so they can't mess with your services
locally, but that's a bit of a stretch. (would need 0666 instead of
0660,right?)

The construct in the SteamVR udev looks intriguing, haven't had a chance to
look it up to see what its effect is.

rpavlik:
This is where that "Move discussion to new issue" button in GitLab would be handy :D  Just spent 5 minutes looking for this, only to discover I had never merged it.

PR #599, opened 2019-05-13
WIP: Create .appveyor.yml

rpavlik:


godbyk:
I like the use of vcpkg!

Here's my [previous attempt](https://github.com/OSVR/OSVR-Core/blob/appveyor/appveyor.yml), in case there's anything worth stealing. Something was causing the build to fail, but I don't recall what it was now. I think I couldn't download some dependency I needed or something.

Some of the other OSVR repositories have appveyor.yml files already, too. However, you might want to update those to make use of vcpkg (and generate packages) if you haven't already.

rpavlik:
yeah, I have a WIP vcpkg for osvr-core but vcpkg isn't very friendly to the idea of submodules with your deps.

PR #601, opened 2019-05-17
Update QuatExpMap.cpp

rpavlik:
Replace M_PI with EIGEN_PI here too. Trying to fix a build.


russell-taylor:
Seems like a good idea; the code depends on Eigen in any case and this provides a standard way to get at PI.

PR #603, opened 2019-06-01
uvbi: Make sure we link against enough opencv libs

rpavlik:
Found this when trying to build it outside of the tree, figured I should backport it.


russell-taylor (approved):
Looks reasonable to me.

